PR #10, opened 2020-04-30
Add timeout when pinging Docker daemon

darrenfoong:
This PR adds a timeout for `cli.Ping()` by using `context.WithTimeout()` to allow users to debug Testcontainer issues more easily.

I had a misconfigured `/etc/docker/daemon.json`, in which I included only `tcp://xxx` in `hosts` while omitting `unix:///var/run/docker.sock`. As a result, Ryuk is constantly stuck in `Ping()`, and Testcontainers fails with "Can not connect to Ryuk". Had there been a timeout and an error message, I might have found the problem earlier, without having to examine the Go source code.

I'm not an experienced Go programmer (started less than 24 hours ago because of this), so I'll be happy to know better ways of dealing with contexts.

### DIFF
--- a/main.go
+++ b/main.go
@@ -24,17 +24,10 @@ func main() {
 
 	cli, err := client.NewEnvClient()
 	if err == nil {
-		ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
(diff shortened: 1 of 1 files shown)


bsideup:
@darrenfoong thanks! I've applied a small change, otherwise looks good :)

PR #11, opened 2020-05-20
Retry prunning on error

### LINKED ISSUE #1: Not allowed to prune networks in paralell
When trying to remove networks, these messages might appear in the logs:
`level=error msg="Handler for POST /v1.29/networks/prune returned error: a prune operation is already running"`

bsideup:
Fixes #1


main.go
```diff
@@ -14,13 +14,14 @@ import (
...
 	"github.com/docker/docker/client"
+	"gopkg.in/matryer/try.v1"
 )
 
 var port = flag.Int("p", 8080, "Port to bind at")
 
 func main() {
 	flag.Parse()
-	log.Printf("Starting on port %d...", *port)
+	log.Println("Pinging Docker...")
```
bsideup, line 24: the logging changes are unrelated a bit but a good general improvement

kiview: I agree, I think it would be even a good idea to log the actual prune operation as well, potentially with the retry number, just as some telemetry?

main.go
```diff
@@ -132,15 +141,29 @@ TimeoutLoop:
 			}
 		}
 
-		networksPruneReport, err := cli.NetworksPrune(context.Background(), args)
-		for _, networkID := range networksPruneReport.NetworksDeleted {
-			deletedNetworks[networkID] = true
-		}
+		try.Do(func(attempt int) (bool, error) {
```
kiview, line 144: This is a completely 3rd party library, or?
So basically just applied [this pattern](https://github.com/matryer/try#delay-between-retries), correct?

main.go
bsideup: Will add 👍

PR #19, opened 2020-11-27
Allow multiple connections

bsideup:


main.go
```diff
@@ -166,7 +176,20 @@ TimeoutLoop:
 			}
 			return shouldRetry, err
 		})
+
+		try.Do(func(attempt int) (bool, error) {
+			imagesPruneReport, err := cli.ImagesPrune(context.Background(), args)
```
rnorth, line 181: Should we prune? 

It feels like this is a bigger action than just cleaning up images that have been used by Testcontainers, since it will remove all untagged images on the system regardless of whether they belong to Testcontainers or not.

bsideup: 1) old code
2) we pass `args` that includes a set of filters that limit it to resources created by the current session only

README.md
```diff
@@ -1,6 +1,6 @@
 # Moby Ryuk
 
-This project helps you to remove containers/networks/volumes by given filter after specified delay.
+This project helps you to remove containers/networks/volumes/images by given filter after specified delay.
```
bsideup, line 3: unrelated change, accidentally committed another contribution 😅 

the image pruning algorithm remains the same

main.go
```diff
@@ -39,21 +40,46 @@ func main() {
 
 	deathNote := make(map[string]bool)
 
-	connected := make(chan bool)
-	disconnected := make(chan bool)
+	firstConnected := make(chan bool, 1)
+	var wg sync.WaitGroup
```
bsideup, line 44: algorithm:

1. create a `WaitGroup`
2. every new connection (handled in a new goroutine) increases it by `1`
3. Every closed connection sleeps for 10s and then decrements `WaitGroup` by `1`
4. The main goroutine waits for `WaitGroup` to become `0` (all connections are closed)

The 10 second delay is "reconnect time", so that if a new connection comes in during this delay, `WaitGroup` will be incremented and the main goroutine will continue waiting

main.go
bsideup: apparently I forgot to pull and the diff was bigger that it was supposed to be, sorry!

rnorth (approved):
LGTM - my previous comments on image pruning should seemingly have been targetting a different commit 😄

PR #21, opened 2020-12-17
Multi-architecture image build using buildx on GitHub Actions

rnorth:
Multi-architecture image build using buildx on GitHub Actions.

Publishing an image for:

- AMD64 (to match status quo)
- ARM64 (for modern ARM servers/Apple M1)
- S390X (for IBM Z Series)
- PPC64LE (for IBM P Series)
- ARMv7 (for Raspberry Pi)

Because the current GitHub-Docker Hub integration does not support multi-arch builds, it is necessary to use GitHub Actions to build the images with Docker buildx (leveraging QEMU emulation).


rnorth:
I've set up a new account on Docker Hub dedicated to builds, `testcontainersbuild`. 

Using a private repo on Hub, `testcontainersbuild/temp-test`, I've done a trial run publication in [this job](https://github.com/testcontainers/moby-ryuk/runs/1572740430?check_suite_focus=true) which results in the following test image being published:

image

Next steps: 

- [ ] reconfigure the workflow to push to `testcontainers/ryuk`, on `release` events only
- [ ] disable the current GitHub-Docker integration
- [ ] review, merge, and do a final test by cutting a new release

.github/workflows/test-docker-image-creation.yml
```diff
@@ -0,0 +1,23 @@
...
+    - name: Docker info
+      run: docker info
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v1
+    - name: Set up Docker Buildx
+      id: buildx
+      uses: docker/setup-buildx-action@v1
+    - name: Buildx inspect
+      run: docker buildx inspect
+    - name: Build the Docker image for multiple architectures without pushing anywhere
```
rnorth, line 22: Change to the `build-push-action`

.github/workflows/publish-docker-image.yml
```diff
@@ -0,0 +1,35 @@
...
+      run: docker buildx inspect
+    - name: Build and push
+      uses: docker/build-push-action@v2
+      with:
+        context: .
+        file: ./Dockerfile
+        platforms: linux/amd64,linux/arm/v7,linux/arm64,linux/ppc64le,linux/s390x
+        push: true
+        tags: |
+          testcontainers/ryuk:latest
```
bsideup, line 34: I would suggest we push only the tag and not `:latest`, to avoid a confusion when there are more changes on `master` but `:latest` still points to the "latest release", but previously our `:latest` was more like `-SNAPSHOT`, on every `master` commit

.github/workflows/test-docker-image-creation.yml
```diff
@@ -0,0 +1,27 @@
+name: Test multi-arch Docker Image creation
```
bsideup, line 1: can't we use GHA's conditions and consolidate both YAML files?

.github/workflows/publish-docker-image.yml
rnorth: Sure, I'm happy not to update `latest` - we should discourage its use.

.github/workflows/test-docker-image-creation.yml
rnorth: Fair point - have done so.

bsideup (approved):
💯

adriancole:
Thanks so much!

PR #22, opened 2021-01-22
Prune on signals SIGINT and SIGTERM

PSanetra:
This PR adds handling of the os signals SIGINT and SIGTERM so that the container will prune immediately after receiving one of them. Currently the process just stops and does not prune at all.

### DIFF
--- a/main.go
+++ b/main.go
@@ -8,8 +8,11 @@ import (
 	"log"
 	"net"
 	"net/url"
+	"os"
+	"os/signal"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/docker/docker/api/types"
@@ -52,10 +55,19 @@ func main() {
 	}
 	log.Println("Received the first connection")
 
-	wg.Wait()
-	log.Println("Timed out waiting for re-connection")
+	signals := make(chan os.Signal, 1)
+	signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
+
+	go func() {
+		wg.Wait()
+		log.Println("Timed out waiting for re-connection")
+		signals <- syscall.SIGTERM
+	}()
+
+	<-signals
 
 	prune(cli, deathNote)
+	os.Exit(0)
 }
 
 func processRequests(deathNote map[string]bool, firstConnected chan<- bool, wg *sync.WaitGroup) {
--- a/main.go
+++ b/main.go
@@ -67,7 +67,6 @@ func main() {
 	<-signals
 
 	prune(cli, deathNote)
-	os.Exit(0)
 }
 
 func processRequests(deathNote map[string]bool, firstConnected chan<- bool, wg *sync.WaitGroup) {


PSanetra:
@bsideup @rnorth Any thoughts?

I would like to have this change as preparation for using ryuk in [HofmeisterAn/dotnet-testcontainers](https://github.com/HofmeisterAn/dotnet-testcontainers).

bsideup:
@PSanetra could you please reduce the changeset to a bare minimum?

PSanetra:
@bsideup I have inlined the `handleConnection` function again and split the commit into two:

* one for making the DeathNote thread safe for concurrent read and write operations
* one for the original purpose of this PR

I am not sure if there is still much to leave out of this PR.

bsideup:
@PSanetra the PR still includes a major refactoring (the addition of `DeathNote`) - can we please do it separately, while focusing this PR on handling the signals? Thanks!

PSanetra:
@bsideup I have extracted the deathnote reafactoring into another PR: https://github.com/testcontainers/moby-ryuk/pull/24

PSanetra:
@bsideup do you have time to have another look at this PR?

lanwen (commented):
Hey @PSanetra , thank you for a great PR, it really improves the tool. However, would you mind adding a few more things?
There are log lines that could be improved, as you anyway changed that.

Also doing 
```diff
--- a/main.go
+++ b/main.go
@@ -20,7 +20,11 @@ import (
        "gopkg.in/matryer/try.v1"
 )
 
-var port = flag.Int("p", 8080, "Port to bind at")
+var (
+       port                  = flag.Int("p", 8080, "Port to bind at")
+       initialConnectTimeout = 1 * time.Minute
+       reconnectionTimeout   = 10 * time.Second
+)
@@ -121,12 +125,12 @@ func waitForPruneCondition(connectionAccepted <-chan 
        select {
-       case <-time.After(1 * time.Minute):
+       case <-time.After(initialConnectTimeout):
                panic("Timed out waiting for the first connection")
        case addr := <-connectionAccepted:
                handleConnectionAccepted(addr)

@@ -141,7 +145,7 @@ func waitForPruneCondition(connectionAccepted <-chan net.Addr, connectionLost <-
        for {
                var noConnectionTimeout <-chan time.Time
                if connectionCount == 0 {
-                       noConnectionTimeout = time.After(10 * time.Second)
+                       noConnectionTimeout = time.After(reconnectionTimeout)
                } else {
                        noConnectionTimeout = never
                }

```

so that we could test it with `main_test.go`

```go
package main

import (
	"net"
	"strings"
	"testing"
	"time"
)

var addr = &net.TCPAddr{
	IP:   net.IPv4zero,
	Port: 5555,
	Zone: "",
}

func init() {
	initialConnectTimeout = 5 * time.Second
	reconnectionTimeout = 1 * time.Second
}

func TestReconnectionTimeout(t *testing.T) {
	acc := make(chan net.Addr)
	lost := make(chan net.Addr)

	done := make(chan struct{})

	go func() {
		waitForPruneCondition(acc, lost)
		done <- struct{}{}
	}()

	acc <- addr
	lost <- addr

	select {
	case <-done:
		return
	case <-time.After(2 * time.Second):
		t.Fail()
	}
}

func TestInitialTimeout(t *testing.T) {
	acc := make(chan net.Addr)
	lost := make(chan net.Addr)

	done := make(chan string)

	go func() {
		defer func() {
			err := recover().(string)
			done <- err
		}()
		waitForPruneCondition(acc, lost)
	}()

	select {
	case p := <-done:
		if !strings.Contains(p, "first connection") {
			t.Fail()
		}
	case <-time.After(7 * time.Second):
		t.Fail()
	}
}
```

PSanetra:
@lanwen Thank you for the review! I have applied the suggested changes.

lanwen (approved):
@bsideup LGTM, I tested it with real tests and tc-java, looks good. Now signals are really handled, triggering prune.

@PSanetra thanks for the great work and a fast update!

lanwen (approved):

bsideup:
Hey @PSanetra!

Thanks for working on it 🎉 
Question: while I appreciate the refactoring, WDYT about reducing the changeset to make it easier to follow and doing the refactoring as a follow up? eg something like this:

```patch
index ce18089..ffd1ee3 100644
--- a/main.go
+++ b/main.go
@@ -8,8 +8,11 @@ import (
 	"log"
 	"net"
 	"net/url"
+	"os"
+	"os/signal"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/docker/docker/api/types"
@@ -52,8 +55,16 @@ func main() {
 	}
 	log.Println("Received the first connection")
 
-	wg.Wait()
-	log.Println("Timed out waiting for re-connection")
+	signals := make(chan os.Signal, 1)
+	signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
+
+	go func() {
+		wg.Wait()
+		log.Println("Timed out waiting for re-connection")
+		signals <- syscall.SIGTERM
+	}()
+
+	<-signals
 
 	prune(cli, deathNote)
 }

```

The reason I am asking is because Ryuk is a critical component and I want to make sure that we are only applying small changes that are easier to follow vs big refactorings, as they are much harder to review.

PSanetra:
@bsideup Yes, I think that could would work, too, but I think it would be good to do this refactoring anyway, as it is currently hard to understand why the 10 seconds timeout works. The refactored code makes this more obvious.

You are right that ryuk is a critical component, but it is not hard to test if it does work as expected as there are not so many test cases. I guess projects, using ryuk will also have some kind of integration tests to check if their integration with ryuk is still working.

bsideup:
@PSanetra I support the idea of doing a refactoring but as a follow up step, to separate functional changes from refactorings 👍

PSanetra:
@bsideup I have moved the original code of this PR to https://github.com/testcontainers/moby-ryuk/pull/33 and pushed your suggested code to this PR.

main.go
```diff
@@ -52,10 +55,19 @@ func main() {
...
+	go func() {
+		wg.Wait()
+		log.Println("Timed out waiting for re-connection")
+		signals <- syscall.SIGTERM
+	}()
+
+	<-signals
 
 	prune(cli, deathNote)
+	os.Exit(0)
```
lanwen, line 70:
```suggestion
```

this would be done anyway 
https://changelog.com/gotime/165

main.go
```diff
@@ -52,10 +55,19 @@ func main() {
 	}
 	log.Println("Received the first connection")
 
-	wg.Wait()
-	log.Println("Timed out waiting for re-connection")
+	signals := make(chan os.Signal, 1)
+	signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
```
lanwen, line 59: for the future refactorings - maybe would be better to just use context? https://pkg.go.dev/os/signal#NotifyContext

that would allow to just close the context instead of pushing the "fake" SIGTERM in the goroutine

lanwen: nice article https://rudderstack.com/blog/implementing-graceful-shutdown-in-go/

lanwen (approved):
with os.Exit removed should be fine to make an iteration

bsideup (approved):
💯

PR #24, opened 2021-03-08
Make DeathNote thread safe

PSanetra:
This PR makes the death note implementation thread safe as it is not safe to read and write on the same map in concurrent go functions: https://stackoverflow.com/a/36167693/4019462

### DIFF
--- a/main.go
+++ b/main.go
@@ -38,12 +38,12 @@ func main() {
 
 	log.Println("Docker daemon is available!")
 
-	deathNote := make(map[string]bool)
+	deathNote := sync.Map{}
 
 	firstConnected := make(chan bool, 1)
 	var wg sync.WaitGroup
 
-	go processRequests(deathNote, firstConnected, &wg)
+	go processRequests(&deathNote, firstConnected, &wg)
 
 	select {
 	case <-time.After(1 * time.Minute):
@@ -55,10 +55,10 @@ func main() {
 	wg.Wait()
 	log.Println("Timed out waiting for re-connection")
 
-	prune(cli, deathNote)
+	prune(cli, &deathNote)
 }
 
-func processRequests(deathNote map[string]bool, firstConnected chan<- bool, wg *sync.WaitGroup) {
+func processRequests(deathNote *sync.Map, firstConnected chan<- bool, wg *sync.WaitGroup) {
 	var once sync.Once
 
 	log.Printf("Starting on port %d...", *port)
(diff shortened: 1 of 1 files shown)


death_note.go
```diff
@@ -0,0 +1,34 @@
+package main
+
+import "sync"
+
+type DeathNote struct {
+	paramSet map[string]bool
+	rwMutex sync.RWMutex
```
lanwen, line 7: can we call this a `lock`?

```suggestion
	lock sync.RWMutex
```

I think that's a common thing to name it like that

death_note.go
```diff
@@ -0,0 +1,34 @@
+package main
+
+import "sync"
+
+type DeathNote struct {
+	paramSet map[string]bool
```
lanwen, line 6:
```suggestion
	notes map[string]bool
```

I think it would be beneficial to stick with the original mythology, then it would be easier to understand the analogy. Then get/read would become `Note(note string)`/`Notes() []string`

death_note.go
```diff
@@ -0,0 +1,34 @@
...
+	rwMutex sync.RWMutex
+}
+
+func NewDeathNote() *DeathNote {
+	return &DeathNote{
+		paramSet: map[string]bool{},
+	}
+}
+
+func (d *DeathNote) AddParam(param string) {
```
lanwen, line 16:
```suggestion
func (dn *DeathNote) AddParam(param string) {
```

completely nit: usually better to use both letter from camel cased type, as it maintains the mind context, easing the read process

lanwen (commented):
@PSanetra awesome to have that now, would be even more to have few basic tests for read/write, as it's a great opportunity to finally have this part tested, as you made it so nicely testable 👍🏻 
I also added few styling/naming suggestions, hopefully, you've seen the anime - it's a great detective story.

PSanetra:
@lanwen I have applied your suggestions and added two tests. Please let me know if you have ideas for additional tests.
Death Note was one of the few animes that I actually watched. I can recommend it, too. :smile:

lanwen (approved):
Looks great for me.

bsideup:
@PSanetra @lanwen but why not https://pkg.go.dev/sync#Map ?

lanwen:
@bsideup oh, totally forgot about it. The only drawback I can see is type assertion on reads, but it should overweight the amount of code saved. 👍🏻

bsideup:
@lanwen I think this is fine given that most of test sessions would only call it once :)

PSanetra:
@bsideup I have replaced that `DeathNote` type with `sync.Map`

lanwen (approved):

bsideup (approved):

PR #25, opened 2021-03-13
Prune all images, not just dangling images

cristianrgreco:
https://github.com/testcontainers/testcontainers-node/issues/189

### DIFF
--- a/main.go
+++ b/main.go
@@ -178,6 +178,7 @@ func prune(cli *client.Client, deathNote map[string]bool) {
 		})
 
 		try.Do(func(attempt int) (bool, error) {
+			args.Add("dangling", "false")
 			imagesPruneReport, err := cli.ImagesPrune(context.Background(), args)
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -178,6 +178,7 @@ func prune(cli *client.Client, deathNote map[string]bool) {
 		})
 
 		try.Do(func(attempt int) (bool, error) {
+			args.Add("dangling", "false")
```
cristianrgreco, line 181: `args` is shared across all prunes (container/network, etc). This only works because we prune images last. I'm not familiar with Go, but ideally what I'd love is to pass a copy of `args` with an extra field, smth like:

```javascript
cli.ImagesPrune(context.Background(),{ ...args, dangling: "false" })
```

Not sure if its possible/if it matters.

bsideup (approved):

bsideup:
@cristianrgreco merged, thank you! 🎉

PR #26, opened 2021-04-16
Fix multi-arch build with pinned versions of buildx/QEMU images

rnorth:
An apparent regression appeared in either buildkit/qemu which was causing the multi-arch build to fail.  #27 describes the problem. 

This PR pins the versions of these dependencies so that we should at least get reproduceable builds. When this PR was first created there was no working combination; however now it looks like this regression was fixed in some intermediate version of `tonistiigi/binfmt`.

### DIFF
--- a/.github/workflows/build-docker-image.yml
+++ b/.github/workflows/build-docker-image.yml
@@ -23,9 +23,13 @@ jobs:
     # Setup for buildx
     - name: Set up QEMU
       uses: docker/setup-qemu-action@v1
+      with: 
+        image: tonistiigi/binfmt:qemu-v6.0.0-12
     - name: Set up Docker Buildx
       id: buildx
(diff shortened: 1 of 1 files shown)


.github/workflows/build-docker-image.yml
```diff
@@ -23,9 +23,13 @@ jobs:
     # Setup for buildx
     - name: Set up QEMU
       uses: docker/setup-qemu-action@v1
+      with: 
+        image: tonistiigi/binfmt:latest
```
bsideup, line 27: Is there a tagged version of it? Or amy other way to pin it?

rnorth: Still WIP! 😆

bsideup (approved):

PR #28, opened 2021-06-13
Upgrade Go, Docker client and Alpine image

reneleonhardt:
* Upgrade Alpine 3.9.6 to 3.13.5
* Upgrade Docker 1.4.2 to 20.10.7
* Upgrade Go 1.13.3 to 1.16.5

Trivy showed several problems in the container image and dependencies, so I upgraded everything.
You can integrate it into GitHub Actions if you want:
https://aquasecurity.github.io/trivy/latest/examples/report/#sarif

PR #30, opened 2021-09-02
Update dependencies

dbyron-sf:
This is a subset of https://github.com/testcontainers/moby-ryuk/pull/28.  Upgrading alpine alone is enough to get `docker scan` to report no vulnerabilities, but upgrading go seems to work as well, so I went for it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ RUN go mod download
 COPY . ./
 RUN make build
 
-FROM alpine:3.9
+FROM alpine:3.13.6
 RUN apk --no-cache add ca-certificates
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,11 +1,11 @@
-FROM golang:1.13.3 as workspace
+FROM golang:1.16.7 as workspace
```
bsideup, line 1:
```suggestion
FROM golang:1.17 as workspace
```

bsideup:
@dbyron-sf it looks like updated Golang can't download the dependencies on `s390x` :(

dbyron-sf:
> @dbyron-sf it looks like updated Golang can't download the dependencies on `s390x` :(

And so we're back to upgrading alpine only, yes?

bsideup (approved):
👍

Issue #18, opened 2020-11-21
Support s390x architecture


codefromthecrypt:
@BbolroC has requested zipkin to support s390x architecture, and we use testcontainers. #15 has notes about arm64 which are relevant here. I noticed that if you use very latest qemu.. it might work, but I can't tell if `go mod download` is slow or hung as it never completes when I run on s390x

adriancole:
@BbolroC can you chase this up? I'm planning to not pay attention as there are other fish to fry

adriancole:
fyi here's where I got on my mac setup with `multiarch/qemu-user-static:5.1.0-7`

```Dockerfile
FROM golang:1.15.4-alpine3.12 as workspace
WORKDIR /go/src/github.com/testcontainers/moby-ryuk
COPY go.mod go.sum ./
RUN go mod download
COPY . ./
RUN CGO_ENABLED=0 go build -i -v -ldflags '-s' -o ./bin/moby-ryuk ./


FROM ghcr.io/openzipkin/alpine:3.12.1
COPY --from=workspace /go/src/github.com/testcontainers/moby-ryuk/bin/moby-ryuk /app
CMD ["/app"]
```


```
$ docker buildx build --progress plain --tag ghcr.io/openzipkin/testcontainers-ryuk:latest --platform=linux/s390x --load .
--snip--
#10 [workspace 4/6] RUN go mod download
#10 1.615 go: github.com/Microsoft/go-winio@v0.4.6: Get "https://proxy.golang.org/github.com/%21microsoft/go-winio/@v/v0.4.6.mod": tls: invalid signature by the server certificate: ECDSA verification failure
#10 ERROR: executor failed running [/bin/sh -c go mod download]: runc did not terminate sucessfully
```

The error is mentioned here golang/go#40949 looks like people avoided packages. I also tried golang 14 same

rnorth:
I've had a quick play today and so far, so good. I wonder if something has been fixed in QEMU in recently - it's hard to know.

I'll aim to tackle this and #15 in the next couple of days.

My draft PR is here if anybody is interested: https://github.com/testcontainers/moby-ryuk/pull/21

rnorth:
This was done in #21

PR #33, opened 2021-12-22
Refactor waiting for prune condition

PSanetra:
As suggested in https://github.com/testcontainers/moby-ryuk/pull/22#issuecomment-999664639 this is a follow up PR to make understanding the way how ryuk waits for the pruning condition easier.


main.go
```diff
@@ -119,11 +115,56 @@ func processRequests(deathNote map[string]bool, firstConnected chan<- bool, wg *
...
+
+	select {
+	case <-time.After(initialConnectTimeout):
+		panic("Timed out waiting for the first connection")
+	case addr := <-connectionAccepted:
+		handleConnectionAccepted(addr)
+	}
+
+	signals := make(chan os.Signal, 1)
+	signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
```
lanwen, line 140: what bothers me, that we listen for signals in a function, not in a main. Also as we refactor that anyways - maybe there is a chance to always handle cleanup as a context close with NotifyContext (even during the initial timeout).
I will try to explain what I think with a code.

PSanetra: There might be trade-offs if there signals are registered earlier, but probably depends on the implementation. Before `signal.Notify` it is still possible to just stop the program via those signal. Only after initialization it is needed to handle those signals explicitly and also not call `signal.Stop` again, as this would make it possible to abort Ryuk while it is trying to cleanup. So I am not sure where you want to cleanup.

lanwen: WDYT of this?

```diff
Index: main.go
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/main.go b/main.go
--- a/main.go	(revision 9bb57be7dd79b35ad1adb24abc015a161a9d7a95)
+++ b/main.go	(date 1643805325365)
@@ -8,7 +8,6 @@
 	"log"
 	"net"
 	"net/url"
-	"os"
 	"os/signal"
 	"strings"
 	"syscall"
@@ -50,8 +49,10 @@
 	connectionLost := make(chan net.Addr)
 
 	go processRequests(deathNote, connectionAccepted, connectionLost)
+	ctx, stop := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
+	defer stop()
 
-	waitForPruneCondition(connectionAccepted, connectionLost)
+	waitForPruneCondition(ctx, connectionAccepted, connectionLost)
 
 	prune(cli, deathNote)
 }
@@ -119,7 +120,7 @@
 	}
 }
 
-func waitForPruneCondition(connectionAccepted <-chan net.Addr, connectionLost <-chan net.Addr) {
+func waitForPruneCondition(ctx context.Context, connectionAccepted <-chan net.Addr, connectionLost <-chan net.Addr) {
 	connectionCount := 0
 	never := make(chan time.Time, 1)
 	defer close(never)
@@ -134,13 +135,10 @@
 		panic("Timed out waiting for the first connection")
 	case addr := <-connectionAccepted:
 		handleConnectionAccepted(addr)
+	case <-ctx.Done():
+		log.Println("Signal received")
+		return
 	}
-
-	signals := make(chan os.Signal, 1)
-	signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
-	// Do not stop handling the signals on returning from this function
-	// as we do not want the pruning process to be interrupted ungracefully.
-	// defer signal.Stop(signals)
 
 	for {
 		var noConnectionTimeout <-chan time.Time
@@ -158,8 +156,8 @@
 			log.Printf("Client disconnected: %s", addr.String())
 			connectionCount--
 			break
-		case sig := <-signals:
-			log.Println("Signal received:", sig)
+		case <-ctx.Done():
+			log.Println("Signal received")
 			return
 		case <-noConnectionTimeout:
 			log.Println("Timed out waiting for re-connection")
```

this way we don't hide the signal handling in the function and it's still explicit in the main

lanwen:
this PR also requires a rebase/merge on master

main.go
PSanetra: @lanwen Looks much better! :+1:

PSanetra:
@lanwen I have applied your suggested changes.
Note: I also had to update Go as `signal.NotifyContext` is only supported in Go >= 1.16.

lanwen (approved):

kiview:
CI is currently failing:
https://github.com/testcontainers/moby-ryuk/runs/5050566673?check_suite_focus=true#step:8:638

This is also not solved by a re-run (at least not the last re-run I did). Is this about a change in the dependencies?

PSanetra:
Looks like the same error as in https://github.com/testcontainers/moby-ryuk/issues/18#issuecomment-731516742. Also related to the s390x architecture build. I am not sure how this was fixed previously. Might be related to the Go version update.

PSanetra:
@kiview I have fixed the s390x build issue and the ci has finished successfully.

kiview (approved):
Thx, LGTM.

@bsideup might want to have a final look before merging.

kiview:
We had the plan to first do the new release before merging this refactoring. Since `0.3.4` has been released, we can therefore merge this change now.

PR #39, opened 2022-07-12
Go updates

IdiosApps:
- `go mod edit "-go=1.17"`
- `go get -u all && go mod tidy` [stackoverflow](https://stackoverflow.com/a/67202563/4261132)
- Add weekly dependabot for gomod updates

> Note that starting from Go 1.17, newly-added indirect dependencies in go.mod are [arranged in a separate require block.](https://stackoverflow.com/questions/69432877/why-there-are-two-require-blocks-in-go-mod-since-go-1-17)

Issue #35, opened 2022-02-01
WCOW Support


PSanetra:
It would be nice to support Ryuk in a windows container environment. Would you be open for such a contribution or do you consider this out-of-scope as the [Java Testcontainers do not support WCOW](https://www.testcontainers.org/supported_docker_environment/windows/#windows-container-on-windows-wcow) at the moment?

It would be nice to have this for the [dotnet-testcontainers](https://github.com/HofmeisterAn/dotnet-testcontainers) project.

kiview:
Hey @PSanetra, sorry for answering late.
 
I think this is a nice idea since a WCOW image of Ryuk would also help testcontainers-java in this regard (missing WCOW Ryuk is a blocker there as well). 

We would be very happy if you could help in adding this support. How do you estimate the effort?

gesellix:
The PR #40 adds a Windows native layer to the existing multi-arch image.

narcis-ro:
Hey guys. Are you still planning to merge either versions? Kind regards

gesellix:
-> https://github.com/testcontainers/moby-ryuk/pull/40#issuecomment-1960321233

gesellix:
A clean pull request is open for discussion at https://github.com/testcontainers/moby-ryuk/pull/110

narcis-ro:
PR was merged. When is it gonna be released?

gesellix:
#120 ist more or less finished. I think after that some dependabot PRs might be merged, too, and then everything should be ready for a release.

gesellix:
Looks like a new release is available: https://github.com/testcontainers/moby-ryuk/releases/tag/0.7.0
:)

mdelapenya:
Indeed, let's reopen this issue in the case a bug in the windows container is found.

Thank you all for your patience 🙏🙇

PR #40, opened 2022-07-30
Add a Windows platform native image (WCOW)

gesellix:
This adds a Windows image to the build, and to the published multi-arch image.

The Windows build ignores the Makefile, though. Please give some feedback if you'd prefer a more future-proof approach. I'd personally suggest to get rid of the Makefile and stick to the Dockerfiles. Another option would be to perform a cross-platform build using that Makefile and publishing the binaries for each platform as a package. Those binaries could then be used in the Dockerfiles.

~I would also suggest to move the Linux Dockerfile to a subdirectory `linux`, so that it becomes more obvious that there are dedicated Dockerfiles for both Linux and Windows.~

~Due to https://github.com/docker/buildx/issues/176 and https://github.com/docker/build-push-action/issues/18 we have to create the Windows image manually, and add it to the multi-arch manifest in another step. I've tested the workflow applied here for one of my own images, see https://github.com/gesellix/go-npipe/releases/tag/v2022-07-31T14-30-00.~

Edit: the current approach is based on https://lippertmarkus.com/2021/11/30/win-multiarch-img-lin/.

Relates to https://github.com/testcontainers/moby-ryuk/issues/35
Relates to https://github.com/testcontainers/testcontainers-java/issues/5621

- [ ] would a dependency update to docker v20.10.17 be valid for everyone, including Linux platforms (see #28)? #39 would also be great to be merged.
- [x] add CI?
- [x] add the Windows platform image to the multi-arch manifest
- [ ] ...?


.github/workflows/build-docker-image.yml
```diff
@@ -22,12 +22,12 @@ jobs:
 
     # Setup for buildx
     - name: Set up QEMU
-      uses: docker/setup-qemu-action@v1
-      with: 
+      uses: docker/setup-qemu-action@v2
+      with:
         image: tonistiigi/binfmt:qemu-v6.0.0-12
```
gesellix, line 27: Is this one required? I'm not sure whether dependabot would notify about updates for that image.

.github/workflows/build-docker-image.yml
```diff
@@ -22,12 +22,12 @@ jobs:
...
     - name: Set up Docker Buildx
       id: buildx
-      uses: docker/setup-buildx-action@v1
+      uses: docker/setup-buildx-action@v2
       with:
         version: v0.6.1
```
gesellix, line 32: Is this one required? I'm not sure whether dependabot would notify about updates for that version.

Dockerfile
```diff
@@ -1,11 +1,11 @@
-FROM golang:1.13.3 as workspace
+FROM golang:1.18 as workspace
```
lanwen, line 1:
```suggestion
FROM golang:1.19 as workspace
```

could be 19 already

windows/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM golang:1.18.4-nanoserver AS workspace
```
lanwen, line 1:
```suggestion
FROM golang:1.19-nanoserver AS workspace
```
was released few days ago

go.mod
```diff
@@ -1,25 +1,23 @@
 module github.com/testcontainers/moby-ryuk
 
-go 1.13
+go 1.18
```
lanwen, line 3:
```suggestion
go 1.19
```

Dockerfile
gesellix: 👍

windows/Dockerfile
gesellix: 👍

go.mod
gesellix: 👍

kiview:
Woah, what are these issues on CI?
https://github.com/testcontainers/moby-ryuk/runs/7725865388?check_suite_focus=true#step:8:596

gesellix:
> Woah, what are these issues on CI?

`local error: tls: bad record MAC` might be a temporary error... very strange. I think retrying the workflows is the best we can do. Not sure if such an error could be an indication of a bad actor tampering with the network. There's a chance that the Golang proxy has issues?

.github/workflows/build-docker-image.yml
gesellix: Version pinning was required for linux/s390x builds, see #26.

.github/workflows/build-docker-image.yml
gesellix: Version pinning was required for linux/s390x builds, see #26.

.github/workflows/build-docker-image.yml
gesellix: I've removed the version pinning, because the linux/s390x builds would fail otherwise... although the version pinning had been introduces due to the same error which appears now 🤪

.github/workflows/build-docker-image.yml
gesellix: I've removed the version pinning, because the linux/s390x builds would fail otherwise... although the version pinning had been introduces due to the same error which appears now 🤪

windows/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM golang:1.19-nanoserver AS workspace
+LABEL builder=true
+
+ENV CGO_ENABLED=0
+#ENV GO111MODULE=on
```
kiview, line 5:
```suggestion
```
Do we need this?

kiview (dismissed):

kiview:
AFAIK GHA uses Cygwin and bash for Windows when executing scripts, so that is something to keep in mind if we want to continue to make use of Make. However, the Makefile is not used anymore after all, or am I missing something?

kiview:
Btw., did you already rebase on #28?

windows/Dockerfile
gesellix: 👍 I've removed the line during rebase/cleanup.

kiview:
We always merge with squash, so don't worry about the actual commit history 🙂 
@bsideup Do you have a need for the Makefile?

windows/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+COPY . ./
+RUN cd /go/src/github.com/testcontainers/moby-ryuk && go get -d \
+ && go vet ./... \
+ && go test ./... \
+ && go build \
+    -v -a \
+    -ldflags "-s -w -extldflags \"-static\"" \
+    -o /bin/moby-ryuk main.go
+
+FROM mcr.microsoft.com/windows/nanoserver:ltsc2022
```
HofmeisterAn, line 18: I think we need to consider different Windows versions. The 2022 base image, is not compatible with every host OS version. This is the [compatibility](https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/version-compatibility?tabs=windows-server-2022%2Cwindows-11-21H2) matrix. Either be build different versions or use 2019 as the base version. It looks like it is compatible with the most recent versions.

.github/workflows/build-docker-image.yml
```diff
@@ -38,9 +38,72 @@ jobs:
...
+        # In case linux/s390x fails with tls issues:
+        # see https://github.com/testcontainers/moby-ryuk/pull/26
+        # and https://github.com/testcontainers/moby-ryuk/pull/40
         platforms: linux/amd64,linux/arm/v7,linux/arm64,linux/ppc64le,linux/s390x
         # Only push if we are publishing a release
         push: ${{ github.event_name == 'release' && github.event.action == 'published' }}
         # Use a 'temp' tag, that won't be pushed, for non-release builds
         tags: testcontainers/ryuk:${{ github.event.release.tag_name || 'temp' }}
+
+  build-windows:
```
HofmeisterAn, line 51: What about a matrix to share the build configuration?

.github/workflows/build-docker-image.yml
```diff
@@ -38,9 +38,72 @@ jobs:
...
+        username: ${{ secrets.DOCKERHUB_USERNAME }}
+        password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+    # Debugging information
+    - name: Docker info
+      run: docker info
+
+    - name: Build image
+      # Use a 'temp' tag, that won't be pushed, for non-release builds
+      run: |
```
HofmeisterAn, line 69: `docker/*` actions instead of the CLI?

.github/workflows/build-docker-image.yml
gesellix: The Windows image requires other build steps, due to https://github.com/docker/build-push-action only working for Linux images. I think it's more maintainable to have a proper job for Windows instead of several `if` clauses in a combined Linux+Windows job.

.github/workflows/build-docker-image.yml
gesellix: I think the docker/build-and-push-action wouldn't work for Windows, see https://github.com/docker/build-push-action/issues/18.

gesellix: In fact, I tried without actually planning: see https://github.com/testcontainers/moby-ryuk/pull/40/commits/54947b0125d55584bc97bd81d3dd6f42f776112e#diff-5b40193e09025fd041de92d38c976d0b469e3ee85bef45a71d05d8b5078b1ee8L66, which failed with an error indicating that buildx wasn't compatible.

HofmeisterAn: Ok, wasn't aware of that. I assumed similar capabilities as the Azure Develops tasks have.

windows/Dockerfile
gesellix: This one might help and even address your other questions below: https://lippertmarkus.com/2021/11/30/win-multiarch-img-lin/

.github/workflows/build-docker-image.yml
gesellix: I need to check whether https://lippertmarkus.com/2021/11/30/win-multiarch-img-lin/ helps.

.github/workflows/build-docker-image.yml
gesellix: I need to check whether https://lippertmarkus.com/2021/11/30/win-multiarch-img-lin/ helps.

windows/Dockerfile
HofmeisterAn: That's interesting. I'll take a look at it tomorrow too, thanks.

gesellix: I gave the cross-build a shot on my Windows system, with LCOW [sic], which failed. So, the "cross-build" is to be taken seriously, or in other words: building Windows images that way works on Linux systems, only. I'm confused now 🤔

Screenshot 2022-08-09 222550

gesellix: Fails on macOS/Docker Desktop, too.
Admittedly, I didn't configure any QEMU or whatever might be required on non-plain Linux systems.

gesellix:
So, I'd like to bring this one forward.

- The mcr.microsoft.com/windows/nanoserver:ltsc2022 base image works on the current GitHub actions Windows runner and runs on recent Windows systems.
- I'm aware that many businesses out there might use older or even out-of-support Windows versions, yet they don't seem to have an issue with the current status quo of Testcontainers not working at all on plain Windows/WCOW.
- I couldn't make the cross-platform build work like described in the linked post. I've tried on Windows (Docker Desktop, LCOW), macOS (Docker Desktop), and plain Linux (Docker engine, Ubuntu 20.04) with recent versions of Docker and _without_ specific updates or tools, because I thrive to keep project setup as simple as possible.
- Containers and Windows are a moving target, so simply taking time and waiting for the tooling around us, might be worth it. See https://github.com/microsoft/windows-host-process-containers-base-image as an example for the ecosystem not being "done", yet.
- I suggest to merge this pull request and continue like that with the other Testcontainers images to have at least a baseline working in modern Windows environments. This would allow you/us to _gather feedback_ and we would have a better chance to decide if supporting older Windows versions would be worth to invest more time and effort. The current pull request doesn't break any existing users.
- I obviously motivated to make it work on _any_ Windows versions, so I'm available to help for any subsequent steps.

What do you think?

kiview:
Sounds completely reasonable to me.

Just one thing, to check that we are on the right track with regards to https://github.com/testcontainers/testcontainers-java/issues/5621:
Did you test using a locally build Ryuk as WCOW in conjunction with Testcontainers for basic WCOW compatibility already? I might also be able to check this out, since I am on a Windows host 🙂

gesellix:
> Did you test using a locally build Ryuk as WCOW in conjunction with Testcontainers for basic WCOW compatibility already?

I didn't try, yet, so yes, would be great if you could give it a shot!

HofmeisterAn:
> Did you test using a locally build Ryuk as WCOW in conjunction with Testcontainers for basic WCOW compatibility already?

In addition, I can do the tests for `tc-dotnet`.

gesellix:
I ran into https://github.com/testcontainers/testcontainers-java/pull/5738 when trying to build testcontainers-java on my Windows machine 😅 
No actual blocker for local tests on WCOW/Ryuk, but it seems like we're discovering other Windows-specific edge cases :)

HofmeisterAn:
I immediately had the problem that I couldn't use the built version with `ltsc2022` (running on 21H2). `ltsc2019` works, though. I still think we should consider `ltsc2019` as the base version. Furthermore, we need to think about something to identify the Windows host version to check the compatibility (with `ltsc2019` probably less important).

Besides that, the image runs smooth on `tc-dotnet`. I only had to add the mount type `npipe` and change the "path" to the Docker socket.

gesellix:
Sounds great, I'm going to give `ltsc2019` a try here on GitHub's Windows runners. We'll certainly have to consider the multi-windows-arch approach or do something manually. Annotations seem to be the best way to stay compatible with different Windows versions). I didn't have enough time to follow every detail of @lippertmarkus' [article](https://lippertmarkus.com/2021/11/30/win-multiarch-img-lin/), yet.

gesellix:
`ltsc2019` builds successfully: https://github.com/gesellix/moby-ryuk/runs/7884751715?check_suite_focus=true#step:5:265

lippertmarkus:
As you use Go I would recommend cross-building the Windows binary on Linux and copy it into a Windows image. Then you don't need a Windows Github runner, it's faster and you can cover more Windows versions than just ltsc2019.
- Example Dockerfile for both Windows and Linux: https://github.com/lippertmarkus/cross-building-windows-and-linux-multi-arch-images/blob/main/windows-and-linux-example/Dockerfile
- Script running buildx to build Windows and Linux images and combine the manifests into a manifest list: https://github.com/lippertmarkus/cross-building-windows-and-linux-multi-arch-images/blob/main/windows-and-linux-example/build.sh

Let me know when you got open questions :)

gesellix:
Thanks for chiming in @lippertmarkus! I actually tried to make your concept work on my local machines, but without success. I just now used your example in a GitHub workflow, which succeeded. It seems like my local machines aren't correctly set up, so I'll postpone to fax that later.

I'm going to update this pull request to support multiple versions of Windows - does anyone like to define the required list of tags or would both `ltsc2019` and `ltsc2022` suffice?

kiview:
I just wanted to test the image in conjunction with Testcontainers until I realized, that I can't use WCOW, since I am currently on Windows 11 Home. So the best way to test and work with WCOW is probably upgrading to Windows 11 Pro, correct?

HofmeisterAn:
> So the best way to test and work with WCOW is probably upgrading to Windows 11 Pro, correct?

I think there are workarounds to enable Hyper-V on Windows 10/11 Home too. But I don't know if it is worth the effort.

Dockerfile
```diff
@@ -1,11 +1,31 @@
...
+CMD ["/moby-ryuk.exe"]
+COPY --from=workspace /bin/moby-ryuk /moby-ryuk.exe
 
-FROM alpine:3.13.6
+FROM alpine:3.16.1 AS linux
 RUN apk --no-cache add ca-certificates
-COPY --from=workspace /go/src/github.com/testcontainers/moby-ryuk/bin/moby-ryuk /app
-CMD ["/app"]
+CMD ["/moby-ryuk"]
+COPY --from=workspace /bin/moby-ryuk /moby-ryuk
```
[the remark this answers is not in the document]
gesellix: Correct!

build-multiarch.sh
```diff
@@ -0,0 +1,54 @@
...
+
+# Get images from Linux manifest list, append and annotate Windows images and overwrite in registry
+docker manifest rm $TARGETIMAGE > /dev/null 2>&1
+lin_images=$(docker manifest inspect $TARGETIMAGE | jq -r '.manifests[].digest')
+
+docker manifest create $TARGETIMAGE $MANIFESTLIST ${lin_images//sha256:/${TARGETIMAGE%%:*}@sha256:}
+
+for VERSION in ${OSVERSIONS[*]}
+do
+  docker manifest rm ${BASE}:${VERSION} > /dev/null 2>&1
```
gesellix, line 43:
```suggestion
  docker manifest rm ${WINBASE}:${VERSION} > /dev/null 2>&1
```

build-multiarch.sh
```diff
@@ -0,0 +1,54 @@
...
+  full_version=`docker manifest inspect ${BASE}:${VERSION} | grep "os.version" | head -n 1 | awk '{print $$2}' | sed 's@.*:@@' | sed 's/"//g'` || true;
```
gesellix, line 44:
```suggestion
  full_version=`docker manifest inspect ${WINBASE}:${VERSION} | grep "os.version" | head -n 1 | awk '{print $$2}' | sed 's@.*:@@' | sed 's/"//g'` || true;
```

gesellix:
I got it working on WCOW, with some modifications necessary in testcontainers-java. So, I guess this one should get a proper review/approval, so we would apply the same pattern to the other Testcontainers images.

One thing to keep in mind: the way the image is being built works for Linux/Buildx - and not on Windows, right? I guess this is a non-issue and might be addressed in the future, in case anyone requires a Windows-native Dockerfile or something like that.

kiview:
To be precise: Building the Linux image on Windows (e.g. even with the default `docker build` command) works just fine, so you were probably talking about building the Windows image on Windows, right? I'd say this is indeed a non-issue 👍

mdelapenya:
Late chiming in here, reading comments fast, and probably missing something, so please don't get me wrong if I come with _"are we aware of https://goreleaser.com/intro/ as the release tool for Go projects?"_

If the root issue is to generate the cross-platform binary (Win, Linux, Mac) and put it into the right Docker image, this could be a straightforward manner to do it.

gesellix:
> you were probably talking about building the Windows image on Windows, right?

Exactly :)

@mdelapenya I know Goreleaser, but wasn't aware of it. So, thanks for chiming in!

Thinking about it, and one might give it a try: would Goreleaser help with our requirement to create multiple variants of a Windows image (for example base images of nanoserver:ltsc2019 and nanoserver:2022)? Goreleaser would certainly be able to create cross-platform binaries from Golang sources, but I don't know if Goreleaser would be able to create images for different Windows versions. I'm not that experienced with the Goreleaser configuration - would you @mdelapenya like to help?

lippertmarkus:
When thinking about it again: you could also use ko, it has Windows container support since some months now and it also works great for Go apps: https://github.com/google/ko#can-i-build-windows-containers

bsideup:
TBH given the complexity of the result, I would really look into using Ko/Goreleaser for this.

@gesellix may I kindly ask you to look into it? Alternatively, we can add it to `main` and then rebase this PR. Ryuk is a critical component of Testcontainers and I want to make sure that we still remember how to build it in 3 months :D

gesellix:
@bsideup sure, I can, it's only a matter of time :)

mdelapenya:
Hi 👋 good news here. I'm working on this PR. I'm giving a try to goreleaser as robutst tool for packaging Go projects, but found a possible limitation with Windows containers (see my comment in https://github.com/goreleaser/goreleaser/discussions/3314#discussioncomment-4277373).

If they are positive in the Windows support for Pro, I'm taking this PR as valid, resolving conflicts and reviewing it.

Thanks for your patience 🙏

gesellix:
@mdelapenya thanks! I had a look at https://ko.build/ (formerly known as google/ko), but I wasn't confident with the experience. It seemed to work when the tooling was based on Golang with their existing multi-platform Docker images, but I personally prefer a more generic build environment. Goreleaser is obviously even more specific and won't fit for other images 😄

mdelapenya:
@gesellix I'm adding all my notes, created while testing this PR locally:

I have this branch locally merged with main, resolving conflicts, and the execution of the script fails locally at the buildx command for linux:

Command:
```
$ ./build-multiarch.sh testcontainers/ryuk:temp no
```

```
ERROR: multiple platforms feature is currently not supported for docker driver. Please switch to a different driver (eg. "docker buildx create --use")
```

The windows build also fails with:
```
 => ERROR exporting to image                                                                                                                                                                                                       0.0s
 => => exporting layers                                                                                                                                                                                                            0.0s
 => => writing image sha256:5e4cab2668ba103d92a5d154ed89399f707c9668206b367a8f7aea8d23af7ebb                                                                                                                                       0.0s
------
 > exporting to image:
------
ERROR: failed to solve: operating system is not supported
```
probably caused by running this on my Mac M1.

After I created a builder including Windows platform, using `docker buildx create --platform windows/amd64 --name windows-builder --use`, the linux execution did work, although I did not see any image loaded into my host:

```
WARNING: No output specified with docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load
no such manifest: docker.io/testcontainers/ryuk:temp
no such manifest: docker.io/testcontainers/ryuk:temp-1809
manifest for image testcontainers/ryuk:temp-1809 does not exist in testcontainers/ryuk:temp
manifest for image testcontainers/ryuk:temp-1903 does not exist in testcontainers/ryuk:temp
manifest for image testcontainers/ryuk:temp-1909 does not exist in testcontainers/ryuk:temp
manifest for image testcontainers/ryuk:temp-ltsc2019 does not exist in testcontainers/ryuk:temp
manifest for image testcontainers/ryuk:temp-2004 does not exist in testcontainers/ryuk:temp
manifest for image testcontainers/ryuk:temp-20H2 does not exist in testcontainers/ryuk:temp
manifest for image testcontainers/ryuk:temp-ltsc2022 does not exist in testcontainers/ryuk:temp
```

If I re-run the command again, updating the script with the `--load` flag to save the images to the current Docker, I get this error for Linux:

```
docker exporter does not currently support exporting manifest lists
```

Therefore, I moved the linux platforms to an array variable and looped through it to get one single `--platform` flag in the `buildx build` command. It makes the load to work and I do see the linux images loaded into my Docker host. But again the manifests are not found 🤷 

I'm going to give a try to https://ko.build. Will ping you here in the issue.

.github/workflows/build-docker-image.yml
```diff
@@ -2,45 +2,41 @@ name: Build multi-arch Docker Image
...
-        push: ${{ github.event_name == 'release' && github.event.action == 'published' }}
+      # In case linux/s390x fails with tls issues:
+      # see https://github.com/testcontainers/moby-ryuk/pull/26
+      # and https://github.com/testcontainers/moby-ryuk/pull/40
+      - name: Cross Build for Linux and Windows
         # Use a 'temp' tag, that won't be pushed, for non-release builds
-        tags: testcontainers/ryuk:${{ github.event.release.tag_name || 'temp' }}
+        run: |
+          ./build-multiarch.sh testcontainers/ryuk:${{ github.event.release.tag_name || 'temp' }} ${{ (github.event_name == 'release' && github.event.action == 'published' && 'yes') || 'no' }}
+...
```
mdelapenya, line 42: Is this line a left-over or put there on purpose?

mdelapenya:
@gesellix more on the code from this branch: I was able to build&push all the tags under my Docker Hub account https://hub.docker.com/r/mdelapenya/ryuk/tags

You can see all the platforms under the same tag:
Screenshot 2022-12-01 at 19 43 39

If you agree, I'd like to push my local branch on top of this one

gesellix:
@mdelapenya sure, feel free to add your changes. Thanks!

build-multiarch.sh
```diff
@@ -0,0 +1,54 @@
...
+# Get images from Linux manifest list, append and annotate Windows images and overwrite in registry
+docker manifest rm $TARGETIMAGE > /dev/null 2>&1
+lin_images=$(docker manifest inspect $TARGETIMAGE | jq -r '.manifests[].digest')
+
+docker manifest create $TARGETIMAGE $MANIFESTLIST ${lin_images//sha256:/${TARGETIMAGE%%:*}@sha256:}
+
+for VERSION in ${OSVERSIONS[*]}
+do
+  docker manifest rm ${WINBASE}:${VERSION} > /dev/null 2>&1
+  full_version=`docker manifest inspect ${WINBASE}:${VERSION} | grep "os.version" | head -n 1 | awk '{print $$2}' | sed 's@.*:@@' | sed 's/"//g'` || true;
```
kuisathaverat, line 44: jq makes the script simpler
```suggestion
  full_version=$(docker manifest inspect ${WINBASE}:${VERSION} |jq -r '.manifests[]|.platform|."os.version"'| sed 's@.*:@@') || true;
```

build-multiarch.sh
```diff
@@ -0,0 +1,54 @@
...
+
+# Get images from Linux manifest list, append and annotate Windows images and overwrite in registry
+docker manifest rm $TARGETIMAGE > /dev/null 2>&1
+lin_images=$(docker manifest inspect $TARGETIMAGE | jq -r '.manifests[].digest')
+
+docker manifest create $TARGETIMAGE $MANIFESTLIST ${lin_images//sha256:/${TARGETIMAGE%%:*}@sha256:}
+
+for VERSION in ${OSVERSIONS[*]}
+do
+  docker manifest rm ${WINBASE}:${VERSION} > /dev/null 2>&1
```
kuisathaverat, line 43:
```suggestion
  # Not sure the remove of the manifest is needed
  docker manifest rm ${WINBASE}:${VERSION} > /dev/null 2>&1
  # if you push the Docker images the manifest is not locally
  docker pull ${WINBASE}:${VERSION}
```

build-multiarch.sh
```diff
@@ -0,0 +1,54 @@
...
+      --pull \
+      --build-arg WINBASE=${WINBASE}:${VERSION} \
+      --target windows \
+      -t "${TARGETIMAGE}-${VERSION}" \
+      .
+    MANIFESTLIST+="${TARGETIMAGE}-${VERSION} "
+done
+
+# Get images from Linux manifest list, append and annotate Windows images and overwrite in registry
+docker manifest rm $TARGETIMAGE > /dev/null 2>&1
```
kuisathaverat, line 36:
```suggestion
# Not sure the remove of the manifest is needed
docker manifest rm $TARGETIMAGE > /dev/null 2>&1
# if you push the Docker images the manifest is not locally
docker pull $TARGETIMAGE
```

.github/workflows/build-docker-image.yml
mdelapenya: I'm not going to be picky about that: we are not implementing an automated, strict YAML validator, which it could be the case at some point. Then, I'd remove them for simplicity of the YAML file

.github/workflows/build-docker-image.yml
```diff
@@ -1,40 +1,63 @@
...
+        run: docker buildx inspect
+
+      # In case linux/s390x fails with tls issues:
+      # see https://github.com/testcontainers/moby-ryuk/pull/26
+      # and https://github.com/testcontainers/moby-ryuk/pull/40
+      - name: Cross Build for Linux and Windows
         # Use a 'temp' tag, that won't be pushed, for non-release builds
-        tags: testcontainers/ryuk:${{ github.event.release.tag_name || 'temp' }}
+        run: |
+          ./build-multiarch.sh testcontainers/ryuk:${{ github.event.release.tag_name || 'temp' }} ${{ (github.event_name == 'release' && github.event.action == 'published' && 'yes') || 'no' }}
```
mdelapenya, line 62: Here I'd suggest passing the commit hash instead of `temp`, as we would be able to even push snapshots based on commits, being possible to actually test them before a release.
```suggestion
          ./build-multiarch.sh "testcontainers/ryuk:${{ github.event.release.tag_name || '${GITHUB_SHA}' }}" "${{ (github.event_name == 'release' && github.event.action == 'published' && 'yes') || 'no' }}"
```

With this change, I'd suggest removing the conditional push and always push, listening for the main branch and tags.

I wonder if it could be useful to also push PRs, using "pr-ID" as tag 🤔 , for the same reason: testing them.

Thoughts?

build-multiarch.sh
```diff
@@ -0,0 +1,63 @@
...
+  ${BUILDX_PUSH} \
+  --pull \
+  --target linux \
+  -t ${TARGETIMAGE} \
+  .
+
+for VERSION in ${OSVERSIONS[*]}
+do
+  echo "Building Windows ${VERSION}"
+    docker buildx build \
```
mdelapenya, line 24: One thing that I noticed in my local workspace is that I needed to create a builder for windows/amd64, with
```
docker buildx create --platform windows/amd64 --name windows-builder --use
```
Otherwise, the windows platform was not available in my build context.

Do you know if the Docker installed in the ubuntu-latest comes with that support?

We could print `docker buildx ls` to verify it

build-multiarch.sh
mdelapenya: Thanks @kuisathaverat for taking a look, much appreciated!

.github/workflows/build-docker-image.yml
kuisathaverat, line 62:
```suggestion
        run: |
          ./build-multiarch.sh testcontainers/ryuk:${RYUK_VERSION} ${IS_RELEASE}
          env:
            RYUK_VERSION: ${{ github.event.release.tag_name || 'temp' }}
            IS_RELEASE: ${{ (github.event_name == 'release' && github.event.action == 'published' && 'yes') || 'no' }}
```

mdelapenya: Indeed, and thanks for participating!

Could you move the `env` block section above run, for readability?

build-multiarch.sh
```diff
@@ -0,0 +1,63 @@
+TARGETIMAGE=${1:-target/image:ci}
+IS_RELEASE=${2:-no}
+WINBASE="mcr.microsoft.com/windows/nanoserver"
+OSVERSIONS=("1809" "1903" "1909" "ltsc2019" "2004" "20H2" "ltsc2022")
```
mdelapenya, line 4: Given the [compatibility matrix](https://learn.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/version-compatibility?tabs=windows-server-2022%2Cwindows-11), suggested by @HofmeisterAn, shows there are two major Windows versions, I suggest keeping this Windows OS array smaller. Wdyt?
```suggestion
OSVERSIONS=("ltsc2019" "ltsc2022")
```

.github/workflows/build-docker-image.yml
gesellix: I'd prefer Git hashes, because that's the handle useful for a checkout and more independent from GitHub.

build-multiarch.sh
HofmeisterAn: It should be fine. The .NET SDK configuartion ships these versions: https://github.com/dotnet/dotnet-docker/tree/main/src/sdk/6.0. `1809` is `ltsc2019`?

.github/workflows/build-docker-image.yml
mdelapenya: We can create `latest` and `$commitSha` tags for commits on the main branch (considering the mutability of the `latest` tag as something already known by the consumers, alongside caching issues when the local latest is outdated), and both `PR-$ID` and `$commitSha` for PRs.

At the end, the storage cost is depreciable and managed by Docker Hub.

gesellix: Yes, I don't have any objections against `PR-$ID` as long as there is a tag with `$commitSha` :)

Regarding the `latest` tag: that not the one being used in Testcontainers, right? As long as Testcontainers itself uses non-mutable tags, I think `latest` can be kept for convenience.

gesellix: All I'm writing here is my personal opinion or experience, because it's up to the maintainers to decide such questions. So please take my answers as input or feedback, and not as decision. :)

Dockerfile
```diff
@@ -1,11 +1,31 @@
-FROM golang:1.18 as workspace
+# dummy value for linux builds
+ARG WINBASE=scratch
```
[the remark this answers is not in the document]
gesellix: I prefer combined Dockerfiles as long as they aren't too complex, because people forget to keep Dockerfiles in sync (only few people seem to care about Windows). Yet, I understand your point of view and it's ok to split them.
I think this is something which can be changed later, back and forth, because the "interface" towards the user is the published image in a registry. So, no hard feelings from my side.

.github/workflows/build-docker-image.yml
mdelapenya: Indeed, we are never going to use latest in the testcontainers libraries. We can provide a latest pointer to the latest commit in the case somebody want to leaves in the edge (and eventually contribute to the upstream if a bug is found)

mdelapenya:
I think this is ready to be merged. It produces wcow containers alongside linux ones (we tested on both Mac and Windows with `docker pull mdelapenya/ryuk:temp3`)

### Mac M1
```
docker run --rm -v /var/run/docker.sock:/var/run/docker.sock --privileged mdelapenya/ryuk:temp3  
2022/12/02 16:51:44 Pinging Docker...
2022/12/02 16:51:44 Docker daemon is available!
2022/12/02 16:51:44 Starting on port 8080...
2022/12/02 16:51:44 Started!
^C2022/12/02 16:51:46 Signal received
2022/12/02 16:51:46 Removed 0 container(s), 0 network(s), 0 volume(s) 0 image(s)
```

### Windows 10
```
docker run --rm -it -v \\.\pipe\docker_engine:\\.\pipe\docker_engine mdelapenya/ryuk:temp3
2022/12/02 17:36:02 Pinging Docker...
2022/12/02 17:36:02 Docker daemon is available!
2022/12/02 17:36:02 Starting on port 8080...
2022/12/02 17:36:02 Started!
```

@gesellix if you agree, I'm updating the description to the latest state of the PR (for discoverability of the real changes), and once approved, we merge and release as `v0.4.0`

.github/workflows/build-docker-image.yml
```diff
@@ -1,40 +1,65 @@
...
-      with:
-        context: .
-        file: ./Dockerfile
-        platforms: linux/amd64,linux/arm/v7,linux/arm64,linux/ppc64le,linux/s390x
-        push: false
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-go@v3
+        with:
+          go-version: '1.18'
```
gesellix, line 18:
```suggestion
          go-version: '1.19'
```

.github/workflows/build-docker-image.yml
```diff
@@ -1,40 +1,65 @@
...
+        run: docker buildx inspect
+
+      # In case linux/s390x fails with tls issues:
+      # see https://github.com/testcontainers/moby-ryuk/pull/26
+      # and https://github.com/testcontainers/moby-ryuk/pull/40
+      - name: Cross Build for Linux and Windows
+        env:
+          RYUK_VERSION: ${{ github.event.release.tag_name || github.sha }}
+          IS_RELEASE: ${{ (github.event_name == 'release' && github.event.action == 'published' && 'yes') || 'no' }}
         # Use a 'temp' tag, that won't be pushed, for non-release builds
```
gesellix, line 63:
```suggestion
```

Dockerfile
```diff
@@ -1,11 +1,25 @@
-FROM golang:1.18 as workspace
+# dummy value for linux builds
+ARG BASE_IMAGE=scratch
+
+FROM --platform=${BUILDPLATFORM} golang:1.18 AS workspace
```
gesellix, line 4:
```suggestion
FROM --platform=${BUILDPLATFORM} golang:1.19 AS workspace
```

Dockerfile.windows
```diff
@@ -0,0 +1,25 @@
+ARG BASE_IMAGE=scracth
+
+# dummy value for linux builds
+FROM --platform=${BUILDPLATFORM} golang:1.18 AS workspace
```
gesellix, line 4:
```suggestion
FROM --platform=${BUILDPLATFORM} golang:1.19 AS workspace
```

Dockerfile.windows
```diff
@@ -0,0 +1,25 @@
+ARG BASE_IMAGE=scracth
+
+# dummy value for linux builds
```
gesellix, line 3:
```suggestion
```

Dockerfile
```diff
@@ -1,11 +1,25 @@
-FROM golang:1.18 as workspace
+# dummy value for linux builds
```
gesellix, line 1:
```suggestion
```

Dockerfile
```diff
@@ -1,11 +1,25 @@
...
+ARG BASE_IMAGE=scratch
```
gesellix, line 2: We might even remove the default `scratch` value?

Dockerfile.windows
```diff
@@ -0,0 +1,25 @@
+ARG BASE_IMAGE=scracth
```
gesellix, line 1: We might even remove the default `scratch` value?

README.md
```diff
@@ -8,7 +8,8 @@ This project helps you to remove containers/networks/volumes/images by given fil
 
         $ ./bin/moby-ryuk -p 8080
         $ # You can also run it with Docker
-        $ docker run -v /var/run/docker.sock:/var/run/docker.sock -p 8080:8080 quay.io/testcontainers/ryuk
+        $ docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -p 8080:8080 testcontainers/ryuk
+        $ docker run --rm -v "//./pipe/docker_engine://./pipe/docker_engine" -p 8080:8080 testcontainers/ryuk
```
gesellix, line 12:
```suggestion
        $ # Example for Windows
        $ docker run --rm -v "//./pipe/docker_engine://./pipe/docker_engine" -p 8080:8080 testcontainers/ryuk
```

gesellix:
> if you agree, I'm updating the description to the latest state of the PR (for discoverability of the real changes), and once approved, we merge and release as v0.4.0

@mdelapenya yes, please, feel free to take over! :)

Dockerfile
mdelapenya: We are still in Go 1.18, so I'd keep it as is, and use a follow-up PR to bump the Go version 🙏

README.md
HofmeisterAn: Quotes are not necessary, maybe just `docker run --rm -v //./pipe/docker_engine://./pipe/docker_engine -p 8080:8080 testcontainers/ryuk`?

Dockerfile.windows
mdelapenya: Done!

Dockerfile
mdelapenya: Done!

README.md
gesellix: Not sure about the quotes. As far as I remember, the `:` triggered some strange behaviour in a shell on Windows. In which shell did you check the examples? I don't remember if I used PowerShell or ConEmu/Git Bash.

HofmeisterAn: Mounting the Unix socket would fail for these shells too. I've tested it on Windows PowerShell and CMD, on macOS Z shell and fish.

.github/workflows/build-docker-image.yml
```diff
@@ -57,8 +60,28 @@ jobs:
...
+        run: docker pull testcontainers/ryuk:${RYUK_VERSION}
+
+      - name: Run Ryuk on Linux
+        if: ${{ matrix.os == 'ubuntu-22.04' }}
+        run: docker run -v /var/run/docker.sock:/var/run/docker.sock testcontainers/ryuk:${RYUK_VERSION}
+
+      - name: Run Ryuk on Windows
+        if: ${{ matrix.os == 'windows-2022' }}
+        run: docker run -v //./pipe/docker_engine://./pipe/docker_engine testcontainers/ryuk:${RYUK_VERSION}
+        continue-on-error: true
```
gesellix, line 87: Is this line still required?

HofmeisterAn: Probably we need to change it slightly, but we would like to keep something like this to test the built image. Due to the complexity and amount of combinations an error sneaked in. The `linux/amd64` and `windows/amd64` contained the ARM binary.

gesellix: Yes, I also think that the checks are required and make sense. I only asked for the "continue on error" line, which is at the very end of the workflow and doesn't have any effect (?).

HofmeisterAn: Yes, you are right. `continue-on-error` is not necessary. We will adjust it with the update that fixes the binaries. Unfortunately, for Windows more changes are necessary due to: https://github.com/actions/runner-images/issues/6688.

mdelapenya:
@gesellix Hi 👋 We are thinking about revisiting this PR in order to get the windows container for Ryuk. We explored the options with goreleaser, although with very few success, as the ability to merge manifests coming from different builders is a Pro feature.

Said that, I think it would be very beneficial if you and I can sync, probably in a zoom call or slack huddle, so we can make this PR into Ryuk. Wdyt?

gesellix:
Hey @mdelapenya :) I'm going to contact you in Slack!

mdelapenya:
We have a really productive online session today, where we were talking on how to unblock this issue after the recent changes in the build system of Ryuk and the failed intents of using goreleaser to build the app without using the Pro version of goreleaser.

As a summary:
- @gesellix shared is use case, where Ryuk needs to run on Windows for certain services he build
- goreleaser has demonstrated to be a powered tool to build Go applications, but without the Pro license it's not possible to split the build for linux/windows and merge them into a single image
- I'm closing this PR as @gesellix will send a fresh new one, using https://github.com/docker-client/echo-server/blob/main/.github/workflows/release.yml as example to build the multi-OS image
- @kiview suggested starting with one Windows version (one of the latest or the most used version), and once we have it working make progress with more versions (2019, 2022...). For that we probably need to update the GH workflow to use a matrix for windows and adjust/create the Dockerfile for Windows, using a build arg for the BASE version.
- @mdelapenya will pair with @gesellix for the final review
- windows users and we are happy 😄 

Therefore, I'm closing this PR waiting for the new one. Thanks all for participating in this healthy discussion 👏

HofmeisterAn:
I would like to mention that our cross-platform build essentially worked, but we encountered issues while publishing the multi-arch manifest to Docker Hub. The pipeline did not merge the manifest, instead it overwrote previously published manifests.

gesellix:
I've restarted to work on the multi-arch manifest: https://github.com/gesellix/moby-ryuk/pull/2

About the goreleaser images and using their manifests: I think that should work somehow, it all comes down to using explicit image references (can you share/link how goreleaser names all the published variants?), collecting their manifests, checking/annotating the `os.version` of the Windows variants, and then merging manifests.

I've tried to extract the "merger" step into a single job, maybe it can be refactored to a custom action with all the images/manifests as _input_ and the desired multi-arch image name as _output_.

HofmeisterAn:
Its been a while, but I think [this](https://github.com/HofmeisterAn/moby-ryuk/compare/main...e8a462a62f892d2dc811f8cce2004e452e71c039) are the latest changes we tested. The names are probably the value of `image_templates`. Does that help?

gesellix:
I'll try (probably not before next week), thanks for the link!

narcis-ro:
Hey guys. Are you still planning to merge either versions? Kind regards

gesellix:
I've rebased the code at https://github.com/gesellix/moby-ryuk/pull/2 to the current upstream (this repo). Let's see, if the workflows still work, which would also be a check, whether the concept stands the test of time :)

narcis-ro:
Thanks for the quick reply @gesellix 
Would be nice to finally have this option for windows

gesellix:
... something seems to be broken on the GitHub's side of Action Runners. I'll check later/tomorrow. Multi-Arch images should appear at https://github.com/gesellix/moby-ryuk/pkgs/container/moby-ryuk (with the older one at https://github.com/gesellix/moby-ryuk/pkgs/container/moby-ryuk/102573888?tag=wip).

gesellix:
See https://github.com/gesellix/moby-ryuk/pkgs/container/moby-ryuk/182452091?tag=20240222.1 with a multi-arch image available for the proof of concept.

Docker image name: `ghcr.io/gesellix/moby-ryuk:20240222.1`

gesellix:
A clean pull request is open for discussion at https://github.com/testcontainers/moby-ryuk/pull/110

narcis-ro:
PR was merged. When is it gonna be released?

PR #45, opened 2022-11-04
chore: bump build system versions

mdelapenya:
## What does this PR do?

- bumps Go to 1.18, using it in the GH action to run the tests
- bumps golang.org/x/sys dependency, as it was using an outdated one, which made it impossible to run the tests
- resolves the duplication of logrus dependency: sirupsen (lowercase) Vs Sirupsen
- runs the tests in the GH action
- adds a make goal for running checks (format, vet, test): `make verify`

## Why is it important?
Go 1.17 reached EOL on Aug 2th, 2022 (see https://endoflife.date/go) for that reason we are bumping it to 1.18.

In order to get gamiliar with the codebase, I tried to run the tests, finding issues with outdated dependencies: sys and sirupsen (both indirect). After fixing them, I noticed the GH action was not running the tests, that's why I added a step in the workflow to verify the project.

## Follows-up
The existing `make lint` is not valid anymore. I'm planning to use golangci-lint as part of [`pre-commit`](https://pre-commit.com)


.github/workflows/build-docker-image.yml
```diff
@@ -20,6 +20,9 @@ jobs:
         username: ${{ secrets.DOCKERHUB_USERNAME }}
         password: ${{ secrets.DOCKERHUB_TOKEN }}
 
+    - name: Verify project
```
eddumelendez, line 23: I think this should be part of another workflow, right? ci.yml maybe? so, PRs are checked instead of doing it when the image is built and publish

mdelapenya: I had that thought too, but I checked this workflow is run on PRs, which is weird to build the image on them

eddumelendez: just saw the published 🤦🏽‍♂️

eddumelendez (approved):

HofmeisterAn (commented):
The `docker/setup-*@v1` actions are outdated in `build-docker-image.yml`. `docker/setup-buildx-action@v1` shows following warning:

> ##[warning]The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
##[warning]The `save-state` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

May not relate to this PR, but should we update them to v2?

mdelapenya:
> May not relate to this PR, but should we update them to v2?

Indeed. I'll add them to this PR as a general bump of versions. I can rename the PR

.github/workflows/build-docker-image.yml
mdelapenya: I'd say that we should move the publish to a different pipeline that runs on tags, only

HofmeisterAn (approved):

.github/workflows/build-docker-image.yml
kiview: Yes, sounds like the simpler approach 👍

PR #47, opened 2022-11-08
chore(ci): only push Docker image on tags

mdelapenya:
## What does this PR do?
It copies the current build pipeline to a push pipeline, removing any interaction to Docker Hub from the build one.

The push pipeline will remove any scripted conditional in the pipeline to detect that the Github event was triggered by a release. This is possible because the workflow will only listen for published releases events.

## Why is it important?
Separation of concerns: we want to build on pushes and PRs, but only publish on tags. It will also increase the maintainability of the pipelines: having separate descriptors will help reviewers to identify where a change must be done.

## Follow ups
We could add snapshots to the push-to-main event. Because the tags will be stored at Docker Hub (no issue for us with the space), we could build&push the Docker image for the main branch, using the last commit as Docker tag. We could use this tag in e2e tests everywhere without needing to "locally" build Ryuk for specific purpose, just as  simple as consuming the "snapshot".

### DIFF
--- a/.github/workflows/publish-docker-image.yml
+++ b/.github/workflows/publish-docker-image.yml
@@ -1,25 +1,32 @@
-name: Build multi-arch Docker Image
+name: Release multi-arch Docker Image
 
 on:
-  push:
-    branches: [main]
-  pull_request:
+  release:
+    types: [published]
 
 jobs:
-  build:
+  release:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
+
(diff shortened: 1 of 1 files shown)


eddumelendez (approved):
LGTM

HofmeisterAn (approved):

kiview:
The follow-up also sounds good. Another idea to consider would be trying out YAML templating, to reuse parts of the pipeline definition (but only if we see even more duplications in the future).

PR #48, opened 2022-11-10
feat: add golangci-lint

hhsnopek:
## Context

This pull request introduces `golangci-lint` into the existing CI/CD Pipeline to increase the project's code health. Applying `go fmt` to all files and fixes a lot of missing error handling.


main.go
```diff
@@ -188,12 +189,15 @@ func prune(cli *client.Client, deathNote *sync.Map) {
...
+				err := cli.ContainerRemove(context.Background(), container.ID, types.ContainerRemoveOptions{RemoveVolumes: true, Force: true})
+				if err != nil {
+					log.Println(err)
+				}
 				deletedContainers[container.ID] = true
 			}
 		}
 
-		try.Do(func(attempt int) (bool, error) {
+		err = try.Do(func(attempt int) (bool, error) {
```
mdelapenya, line 200: A probably more robust approach would be extracting the try block to a function where it receives a closure. That way we could reuse it across the three different try blocks Wdyt?

hhsnopek: Agreed but I think that is out-of-scope for this pull request and could be done in a follow-up PR. thoughts?

kiview: I agree, let's make sure this PR only focuses on the direct `go fmt` changes and linting errors.

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,49 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-go@v3
+        with:
+          go-version: 1.19
+      - uses: actions/checkout@v3
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@07db5389c99593f11ad7b44463c2d4233066a9b1
+        with:
+          # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
```
kiview, line 27: I think we should remove all `Optional` keys, since we would likely lookup the docs if we ever want to reconfigure.

mdelapenya: Indeed. We added them to testcontainers-go in https://github.com/testcontainers/testcontainers-go/issues/569, but we should remove it to clean up the file

main.go
```diff
@@ -188,12 +189,15 @@ func prune(cli *client.Client, deathNote *sync.Map) {
 			log.Println(err)
 		} else {
 			for _, container := range containers {
-				cli.ContainerRemove(context.Background(), container.ID, types.ContainerRemoveOptions{RemoveVolumes: true, Force: true})
+				err := cli.ContainerRemove(context.Background(), container.ID, types.ContainerRemoveOptions{RemoveVolumes: true, Force: true})
+				if err != nil {
+					log.Println(err)
```
mdelapenya, line 194: Do we want to `continue` here to avoid incrementing the number of deleted containers?

kiview: If I am nitpicking, I would probably add this in a follow-up bugfix PR, to have a slightly better Git history 🙂

mdelapenya: Mmmm I think it is actually part of the change: handle the error, which was not handled in the past

kiview: Which number of deleted containers we would have before? Probably no output, since it would fail? Or we would have the wrong number?

mdelapenya: The error was swallowed, the execution continued adding a container to the map.

Being strict, it could be considered a fix, yeah

kiview: But we can also be pragmatic and fix it here. I think having the discussion was more important than the detail where we fix it (as long as we fix it).

hhsnopek:
@mdelapenya @kiview I apologize for the force push, but I've resolved the comments which means I've removed the additional error handling/logs in-favor of keeping the scope to adding golangci-lint.

main.go
hhsnopek: I'm unsure what the expected functionality should be at this time, happy to discuss it further in another PR as there are a handful of unhandled errors.

kiview (approved):

mdelapenya (approved):

PR #49, opened 2022-11-26
feat: add $RYUK_CONNECTION_TIMEOUT to configure connection timeout

hhsnopek:
## Context

This pull request adds the ability to configure ryku's connection timeout by either a `flag` or environment variable `RYKU_CONNECTION_TIMEOUT` and if both are defined the flag is preferred. The goal of this change is to provide a way to define the reaper's timeout when environments take longer than 60 seconds to initialize.


main_test.go
```diff
@@ -80,7 +87,7 @@ func TestInitialTimeout(t *testing.T) {
 }
 
 func TestPrune(t *testing.T) {
-	cli, err := client.NewClientWithOpts()
+	cli, err := client.NewClientWithOpts(client.FromEnv)
```
hhsnopek, line 90: `client.FromEnv` was added so that Docker client configuration consumes docker configuration by environment variables.

main_test.go
```diff
@@ -91,6 +98,7 @@ func TestPrune(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	cli.NegotiateAPIVersion(context.Background())
```
hhsnopek, line 101: I'm unsure if this is needed, but with this and the previous comment/snippet `client.FromEnv` I was successfully able to run the tests with Colima locally.

main.go
```diff
@@ -20,20 +21,78 @@ import (
...
+}
+
+// newConfig parses command line flags and returns a parsed config. config.timeout
+// can be set by environment variable, RKYU_CONNECTION_TIMEOUT. If both flag and
+// environment variable are defined the flag is prioritized. If an error occurs
+// while parsing RKYU_CONNECTION_TIMEOUT the error is returned.
+func newConfig(args []string) (*config, error) {
+	cfg := config{ReconnectionTimeout: 10 * time.Second}
+
+	fs := flag.NewFlagSet("ryku", flag.PanicOnError)
```
hhsnopek, line 47: should the FlagSet name be `ryku` or `moby-ryku`?

kiview:
Thanks @hhsnopek, in which context did you have the need to increase the Ryuk timeout?

We added the same to Testcontainers for Java some years ago, so I generally understand the need for it:

- https://github.com/testcontainers/testcontainers-java/issues/621
- https://github.com/testcontainers/testcontainers-java/pull/635

In Java, we are using our usual configuration mechanism, which ENV, `~/.testcontainers.properties` and `classpath:testcontainers.properties`, in that order. In know we also have `~/.testcontainers.properties` in Go, is it already hooked up in this case (sorry for asking, I am not super familiar with the codebase, to easily see it from the diff)? And is a flag something idiomatic to our Go implementation?

main_test.go
kiview: I think this should come in its own PR. As already hinted by you in Slack, probably a good idea to have a discussion of the configuration subsystem (/cc @mdelapenya).

main.go
kiview: We use `ryuk.container.timeout` as the property in Java.

bsideup:
FTR it is "ryuk", not "ryku" 😅

Also, I am not sure we need flags here, as Ryuk isn't intended to be used as a standalone CLI tool, so I would focus on environment variables only 👍

main_test.go
hhsnopek: This configures the Docker client for Go package to reference the environment variables when loading the client. In deployed images, like the container it lives in, it shouldn't affect anything. However in a local/test environment, the client won't recognize the Docker context or environment configurations of `DOCKER_HOST` which eliminates the ability to run tests locally.

main.go
hhsnopek: This references the name of the program which called by the command line. In `main` it shows the following:
```
$ ./bin/moby-ryuk -h
Usage of ./bin/moby-ryuk:
  -p int
        Port to bind at (default 8080)
```

With the flagset:
```
$ ./bin/moby-ryuk -h
Usage of ryuk:
  -p int
        Port to bind at (default 8080)
  -timeout duration
        Connection timeout (default 1m0s)
```

hhsnopek:
@kiview My current use cases are:
* debugging Colima issues
* ensuring that the ryuk doesn't timeout before the Strategy.Deadline/Timeout timeout occurs

Go processes are typically configured by flags and environment variables, like Docker's convention where you can use environment variables, flags, and local/global configuration (toml, yaml, json) files. I often see environment variables with a namespace prefix, like `DOCKER_` or `PG`, which matches the `RYUK_` prefix in the PR.

@bsideup since Ryuk runs in a docker container and could be configured by an environment variable and a flag passed in `cmd`, I would suggest keeping both options open. With both exposed the interface/configuration later exposed in `testcontainers-go` via `reaper.go` can also override either programmatically. The order of priority would be: default (60s), environment variable, flag, `testcontainers-go` definition which would match Java's configuration prioritization.

main_test.go
kiview: Please ignore my comment, I was thinking in the context of `tc-go` rather than Ryuk.

hhsnopek:
I used the flag in both testing and debugging by updating my local docker image to have a static flag in [`CMD`](https://github.com/testcontainers/moby-ryuk/blob/main/Dockerfile#L11). The environment variable could have also been used. To be honest, Ryuk can just use the environment variable - this addition alone does nothing until the environment variable is forwarded to the container in testcontainers-go

I'm happy to remove the flag, I don't think Go has an idiom for this

kiview:
Sounds good, then let's just minimize the PR and the implementation. WDYT @mdelapenya?

main.go
```diff
@@ -20,20 +21,70 @@ import (
...
+	err := fs.Parse(args)
+	if err != nil {
+		return nil, err
+	}
+
+	if timeout, ok := os.LookupEnv(connectionTimeoutEnv); ok {
+		parsedTimeout, err := time.ParseDuration(timeout)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse \"%s\": %s", connectionTimeoutEnv, err)
+		}
```
mdelapenya, line 63: Question about this error handling: do we want to fail the execution when the env var value is not a duration or fallback to the default connection timeout?

mdelapenya: Because we have a default, let's fall back to it if possible.

I think that once we have that, this is ready to go!

kiview: I think failing is the more Unix-style of doing it. Also, we have to think through that the user for this variable are the Testcontainers implementations. I assume we expect them to always submit a valid duration (unless they would forward user config without any kind of validation). 

So, both approaches have a pro and cons: Either fail more explicitly or be more robust and use a solid fallback. I can live with both 🙂

hhsnopek:
@mdelapenya is there any pending changes required before merging? I believe all feedback has been addressed, please let me know if I have not.

kiview (dismissed):

kiview:
@hhsnopek LGTM. Given the fact that @mdelapenya currently takes a bigger responsibility in the maintenance and development of Ryuk and given https://github.com/testcontainers/testcontainers-go/discussions/653, is it ok for you if we wait until @mdelapenya is back?

mdelapenya (approved):
LGTM, thanks for your patience during the review 🙏

mdelapenya:
BTW, I've just refreshed myself with the comments, specially https://github.com/testcontainers/moby-ryuk/pull/49#issuecomment-1328236103 and https://github.com/testcontainers/moby-ryuk/pull/49#issuecomment-1329412380.

Let's remove the flag support and keep the env variables only, as suggested.

mdelapenya (changes requested):
BTW, I've just refreshed myself with the comments, specially https://github.com/testcontainers/moby-ryuk/pull/49#issuecomment-1328236103 and https://github.com/testcontainers/moby-ryuk/pull/49#issuecomment-1329412380.

Let's remove the flag support and keep the env variables only, as suggested.

mdelapenya:
I'm revisiting this PR after being back from leave. As mentioned above, we are going to focus on env vars for the connection timeout config. But then, as a consequence, we should remove the already existing flag for the port in which Ryuk is listening, to be consistent while configuring its execution.

_The flag for the port already exists_:
```
Usage of ryuk:
  -p int
        Port to bind at (default 8080)
```

PR #50, opened 2022-11-29
Improve CI process & refactor Dockerfile

hhsnopek:
## Context
This pull request splits the CI task from the Docker build task, refactors the Dockerfile to build faster from ~126s to ~58s, and upgrades the alpine-linux version to 3.16. This will increase the feedback loop of tests in the build pipeline for contributors.


.github/workflows/build-docker-image.yml
```diff
@@ -15,9 +15,6 @@ jobs:
         go-version: '1.18'
     - run: go version
 
-    - name: Verify project
```
mdelapenya, line 18: I understand that having multiple, smaller GH workflows is probably more maintainable than having a maybe longer pipeline.

But I do believe that we should not build any artifact if the tests fail, that's why I would keep this pipeline as is, or at least make them dependent, with this condition: package the artifacts (docker build) if and only if the tests pass.

Dockerfile
```diff
@@ -1,11 +1,28 @@
...
+# Install source deps
 COPY go.mod go.sum ./
 RUN go mod download
-COPY . ./
-RUN make build
 
-FROM alpine:3.13.6
+# Copy source & build
+COPY . .
+RUN go build -v -ldflags '-s' -o /bin/ryuk
```
mdelapenya, line 17: I noticed here and [here](https://github.com/testcontainers/moby-ryuk/pull/50/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR19-R30) that the existing make file is not used any more in favor of the go commands.

One thing that I expect from a build system such as Make, and from a CI, is that the CI must use the very same commands as the build system in order to provide developers a way for reproducing a CI build error.

In this PR we are removing the usages of the Makefile in favor of go test, which is fine to me at the moment we are making them explicit in the GH action and any developer would be able to read the GH workflow file.

On the other hand, Make provides an abstraction of the commands which is handy for developers to run the build system.

Which option do we prefer: Make Vs go commands? I would prefer having Make as entry point to the build, but I do not have a strong opinion on directly using go commands, producing a workflow with a simple set of steps.

Wdyt?

kiview: Being more of an outsider I never really understood why Go projects use Make in an extremely simplistic fashion, often having a 1:1 mapping between goal and go command execution. However, if Make tends to be the default build system in the Go community it's of course good to follow this as a best practice and then I agree with @mdelapenya, that it should we used in all corresponding places to create artifacts.

mdelapenya: I'm not seeing Make as a default build system, but as best effort to define a proper build life cycle: compile > test > package as a minimum, probably biased by my previous experience with bigger Go projects.

I'm OK with removing it from the project, or keeping it, but being aware of the implications of it.

hhsnopek: I find make commands which are 1:1 mappings of the go commands are redundant unless standardized between repositories. I removed it from the Dockerfile to reduce the build cycle and explicitly define what's used inside the container. The dockerfile was a bit confusing since the context switched between cli, dockerfile, make stretching the context between three areas in a very short space:

* docker build -t testcontainers/ryuk
* Step into Dockerfile context 
* Step into Make context 
* Step back to Docker context

I can revert this change if you'd like

.github/workflows/build-docker-image.yml
hhsnopek: If it's okay to make it dependent I would prefer that as it's more visible in the checks and runs without the 30m build time. Let me know what you'd prefer

Dockerfile
mdelapenya: I'm fine removing Make :)

.github/workflows/build-docker-image.yml
eddumelendez: I agree with @mdelapenya. It can be just another job in the same workflow

```yml
jobs:
  build:
  build-image:
```

and it will be visible in the checks

Makefile
```diff
@@ -1,40 +0,0 @@
...
-
-build: vet fmt compile
-
-fmt:
-	go fmt $(SOURCE_FOLDER)/...
-
-vet:
-	go vet $(SOURCE_FOLDER)/...
-
-lint:
```
mdelapenya, line 34: @hhsnopek sorry for taking too long to review this PR 🙏  I was focused more on tc-go releases.

One question regarding this change: because we already have a workflow for golangci-lint, is it safe removing this build commands?

Dockerfile
```diff
@@ -1,12 +1,29 @@
...
+# Distributed Image
+# -----------------
+FROM alpine:3.16
+
 RUN apk --no-cache add ca-certificates
-COPY --from=workspace /go/src/github.com/testcontainers/moby-ryuk/bin/moby-ryuk /app
-CMD ["/app"]
+
+COPY --from=build /bin/ryuk /bin/ryuk
+CMD /bin/ryuk
```
HofmeisterAn, line 28:
```suggestion
CMD ["/bin/ryuk"]
```

We can use the exec form, right?

.dockerignore
```diff
@@ -1,2 +1,10 @@
...
+.git/
+.github/
+bin/
+testresources/
+.gitignore
+.golangci.yml
+Dockerfile
+Makefile
+README.md
+main_test.go
```
HofmeisterAn, line 10:
```suggestion
.git/
.github/
bin/
testresources/
.gitignore
.golangci.yml
Dockerfile
main_test.go
README.md
```

Makefile gets removed below.

mdelapenya (approved):

mdelapenya:
Thanks everybody for this review, I'm merging it now

Makefile
mdelapenya: We can add them if needed in a follow-up. Thanks!

PR #52, opened 2022-12-09
feat: build Ryuk using goreleaser

mdelapenya:
## What is this PR doing?
It adds [goreleaser](https://goreleaser.com/) as build tool to cross-build Ryuk in all supported platforms:

#### Binaries
- linux/amd64
- linux/arm64
- linux/ppc64le
- linux/s390x
- darwin/amd64
- darwin/arm64
- windows/amd64

#### Dockers
- linux/amd64
- linux/arm64
- linux/armv7
- linux/ppc64le
- linux/s390x
- windows/amd64 2019lts
- windows/amd64 2022lts
- latest and the current tagged version will be multi-arch

For that, we will need to build each platform on its own GH worker, and finally create the single multi-platform manifest in the very same worker that created the image. For that, an initial stage pushing the first manifest must be done beforehand.

```mermaid
graph TD
    A[Build Ryuk] -->|ubuntu-22.04| B(Build linux/amd64)
    B --> |ubuntu-22.04| C(Create manifest: linux/amd64)
    D[GH Workers] --> |ubuntu-22.04| E[Rest of linux images]
    C --> |Cross-Build| D
    D -->|windows-2019| F[lts2019]
    D -->|windows-2022| G[lts2022]
    E --> H(Annotate the manifest)
    F --> H
    G --> H
  ```

As seen in the diagram, we need to first create a manifest (build+push), i.e. for linux/amd64, in order to create an initial image where to append the rest of manifests in parallel, and once it finishes, build and annotate the rest of the linux and the windows versions.

Each OS-stage will have its own goreleaser config file, so that it knows what to build

## Why is this important?
Cross-build Ryuk in a simple, consistent and standard way

### CODE REVIEW

File: Dockerfile [hunk 1]
```diff
@@ -1,11 +1,6 @@
-FROM golang:1.18 as workspace
-WORKDIR /go/src/github.com/testcontainers/moby-ryuk
-COPY go.mod go.sum ./
-RUN go mod download
-COPY . ./
-RUN make build
-
-FROM alpine:3.13.6
+FROM alpine:3.16.1
 RUN apk --no-cache add ca-certificates
```
HofmeisterAn, line 2: Do we require this line anymore?

Dockerfile.windows
```diff
@@ -0,0 +1,8 @@
+ARG WIN_VERSION
+
+FROM mcr.microsoft.com/windows/nanoserver:${WIN_VERSION}
+
+CMD ["/ryuk.exe"]
+ENTRYPOINT ["/ryuk.exe"]
+
+COPY ryuk.exe /
```
HofmeisterAn, line 8:
```suggestion
ARG WIN_VERSION

FROM mcr.microsoft.com/windows/nanoserver:${WIN_VERSION}

ENTRYPOINT ["/ryuk.exe"]

COPY ryuk.exe /
```

.github/workflows/goreleaser.yml
```diff
@@ -0,0 +1,79 @@
+# .github/workflows/goreleaser.yml
+name: goreleaser
+
+on:
+  push:
+    tags:
+      - '*'
+
+permissions:
+  contents: write
```
HofmeisterAn, line 10: I just want to double-check, why do we require write permissions?

.github/workflows/goreleaser.yml
```diff
@@ -0,0 +1,79 @@
...
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.18
```
HofmeisterAn, line 39: Share Go version as variable?

.github/workflows/goreleaser.yml
```diff
@@ -0,0 +1,79 @@
...
+
+jobs:
+  goreleaser:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-22.04, windows-2019, windows-2022]
+    runs-on: ${{ matrix.os }}
+    env:
+      DOCKER_CLI_EXPERIMENTAL: "enabled"
```
HofmeisterAn, line 20: You can set the env variable global for all jobs.

.github/workflows/goreleaser.yml
```diff
@@ -0,0 +1,79 @@
...
+
+permissions:
+  contents: write
+
+jobs:
+  goreleaser:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-22.04, windows-2019, windows-2022]
```
HofmeisterAn, line 17: Dunno, if we need to think about it, but what happens if one OS fails to push the image? We clean it up afterwards by hand?

HofmeisterAn (commented):
Can we pass arguments to the GoReleaser YML files?

Dockerfile
mdelapenya: I'd say the ca-certificates were present in the old base image, probably to allow SSL connections, so not sure if we should remove the package now

Dockerfile.windows
mdelapenya: Makes sense, let's remove the CMD, as it's not adding anything different from the entrypoint

.github/workflows/goreleaser.yml
mdelapenya: This is something to evaluate, and probably have a rescue condition: if any of the stages fail (create the initial linux/amd image > create the rest of linux | create windows2019 | create windows2022), should we remove the image from the registry? IIRC it was super hard in the past to remove an image from a registry using CLIs/APIs, not sure if nowadays it's possible

Dockerfile
HofmeisterAn: Hm, I do not understand which SSL connections that could be. TC connects to the plain socket and the Docker build stage does not fetch any data.

mdelapenya: TBH I'm not sure why it was there before these changes. If you agree, let's keep it here for this PR and create a follow-up discussion on the real need for the ca-certificates package.

.github/workflows/goreleaser.yml
mdelapenya: From their docs: https://goreleaser.com/ci/actions/#token-permissions

> - contents: write if you wish to
>   - [upload archives as GitHub Releases](https://goreleaser.com/customization/release/), or
>   - publish to [Homebrew](https://goreleaser.com/customization/homebrew/), or [Scoop](https://goreleaser.com/customization/scoop/) (assuming it's part of the same repository)
> - or just contents: read if you don't need any of the above
> - packages: write if you [push Docker images](https://goreleaser.com/customization/docker/) to GitHub
> - issues: write if you use [milestone closing capability](https://goreleaser.com/customization/milestone/)

Probably a good point to discuss if we need to attach the resulting binaries to the Github release 🤔

Dockerfile
HofmeisterAn: OC, sure.

.github/workflows/goreleaser.yml
HofmeisterAn: I think it is fine 👍. I just want to double-check. Pipeline just runs on new tags.

mdelapenya:
Checking with the authors of goreleaser here: https://github.com/goreleaser/goreleaser/discussions/3314#discussioncomment-4790678

mdelapenya:
We are going to test this like this:

- create a baseline branch, `goreleaser`
- merge this PR on that baseline branch
- use a fork (e.g. @mdelapenya) to update the manifests to push the docker images to a different Docker namespace
- update the goreleaser GH action to listen to pushes in the goreleaser branch to test incremental changes

PR #53, opened 2023-01-03
Prevent Ryuk from removing itself

matthewmcnew:
- Workaround for https://github.com/testcontainers/testcontainers-go/issues/698
 - Add generic label to the ryuk image `org.testcontainers.ryuk` which enables the ryuk to detect itself and prevent deletions.

### DIFF
--- a/main.go
+++ b/main.go
@@ -20,6 +20,8 @@ import (
 	"gopkg.in/matryer/try.v1"
 )
 
+const ryukLabel = "org.testcontainers.ryuk"
+
 var (
 	port                  = flag.Int("p", 8080, "Port to bind at")
 	initialConnectTimeout = 1 * time.Minute
@@ -192,7 +194,7 @@ func prune(cli *client.Client, deathNote *sync.Map) (deletedContainers int, dele
 			log.Println(err)
 		} else {
 			for _, container := range containers {
-				_, isReaper := container.Labels["org.testcontainers.ryuk"]
+				_, isReaper := container.Labels[ryukLabel]
 				if isReaper {
 					continue
 				}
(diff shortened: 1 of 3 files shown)


bsideup:
Actually, Ryuk shouldn't be labeled with the same label that it targets, that's a bug in Ryuk's consumer.

Here is the relevant piece in Testcontainers for Java:
https://github.com/testcontainers/testcontainers-java/blob/fa56f24dc77da80c3a77bb14f096c955c93e9e14/core/src/main/java/org/testcontainers/utility/RyukResourceReaper.java#L62-L70

matthewmcnew:
@bsideup Yes, It is a bug in [testcontainers-go](https://github.com/testcontainers/testcontainers-go/issues/698). In the meantime this PR to ensure that ryuk handles this incorrect consumer behavior [as suggested by](https://github.com/testcontainers/testcontainers-go/issues/698#issuecomment-1357430101) @HofmeisterAn.

HofmeisterAn:
Yup, I agree it is in the first place an issue in Ryuk's consumer, but it sounds reasonable to prevent others from running into the same issue. My idea was to start a discussion in `moby-ryuk` first while we are fixing the Go implementation.

main.go
```diff
@@ -192,6 +192,11 @@ func prune(cli *client.Client, deathNote *sync.Map) (deletedContainers int, dele
 			log.Println(err)
 		} else {
 			for _, container := range containers {
+				_, isReaper := container.Labels["org.testcontainers.ryuk"]
```
mdelapenya, line 195: Could we move the string label to a constant and use it here?

mdelapenya: @matthewmcnew done at [5a31bc5](https://github.com/testcontainers/moby-ryuk/pull/53/commits/5a31bc5aaa7d7d4c8d75117015e47e7c7c28214a)

mdelapenya:
@matthewmcnew thanks for the issue report and your work here. Much appreciated!

PR #64, opened 2023-03-07
feat: set ryuk port using the environment variable

mdelapenya:
## What does this PR do?

It allows setting the Ryuk port from the RYUK_PORT environment variable, which takes precedence over the current port flag.

At the same time, we are adding a deprecation warning in the usage of the current port flag, to advertise of the changes. At this moment, if both are set, the environment will take precedence.

Besides that, we are also supporting setting the reconnection timeout using the RYUK_RECONNECTION_TIMEOUT env var.

## Why is it important?
Consistency across the different settings, as we introduced a way to configure Ryuk with env vars instead of with flags.

### DIFF
--- a/main.go
+++ b/main.go
@@ -53,7 +53,7 @@ func newConfig(args []string) (*config, error) {
 	fs := flag.NewFlagSet("ryuk", flag.ExitOnError)
 	fs.SetOutput(os.Stdout)
 
-	fs.IntVar(&cfg.Port, "p", 8080, "Port to bind at")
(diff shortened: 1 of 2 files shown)


kiview:
If we deprecate the usage of flags, we should update the `Usage` section in the README as well, to not use them.

mdelapenya:
> If we deprecate the usage of flags, we should update the `Usage` section in the README as well, to not use them.

Indeed. I added the new env vars but the flag is still there. Updating it now

HofmeisterAn (approved):

eddumelendez (approved):

thomaseizinger (approved):
Rust testcontainers doesn't use this so no objections from my end!

Thanks for the ping!

PR #69, opened 2023-03-27
Update workflow to publish in Google Artifact Registry

eddumelendez:
* Update `docker/login-action` version to `v2`
* Configure the registry
* Update image

### DIFF
--- a/.github/workflows/publish-docker-image.yml
+++ b/.github/workflows/publish-docker-image.yml
@@ -16,10 +16,11 @@ jobs:
     - run: go version
 
     - name: Login to Docker Hub
-      uses: docker/login-action@v1
+      uses: docker/login-action@v2
       with:
-        username: ${{ secrets.DOCKERHUB_USERNAME }}
-        password: ${{ secrets.DOCKERHUB_TOKEN }}
+        registry: us-docker.pkg.dev
+        username: _json_key_base64
+        password: ${{ secrets.TC_GAR_JSON_KEY }}
 
     - name: go-mod verify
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-docker-image.yml
```diff
@@ -55,4 +56,4 @@ jobs:
         platforms: linux/amd64,linux/arm/v7,linux/arm64,linux/ppc64le,linux/s390x
         # Only push if we are publishing a release
         push: true
-        tags: testcontainers/ryuk:${{ github.event.release.tag_name }}
+        tags: us-docker.pkg.dev/testcontainers-oss/testcontainers/ryuk:${{ github.event.release.tag_name }}
```
mdelapenya, line 59: Should we tag both or only the new one?

HofmeisterAn: I am fine pushing older versions manual. Anything else would be a bit difficult (recreating the release), or do you have anything specific in mind? Is `0.4.0` already used in other language implementations? .NET still use `0.3.4` in case developers have issues with the new version it might be a good idea to push the `0.3.4` too.

eddumelendez: existing images were moved manually using [crane](https://github.com/google/go-containerregistry/blob/main/cmd/crane/doc/crane_copy.md)

mdelapenya (approved):

kiview:
Closed, since we currently don't plan to pushing into GAR anymore.

PR #71, opened 2023-03-30
chore: super simple verbose logs when pruning

mdelapenya:
## What does this PR do?
It supports configuring Ryuk to add verbosity when calling the Docker APIs. It can be set with the `RYUK_VERBOSE` environment variable, which accepts boolean values.

Everytime the Docker client is used to list/prune resources (containers, networks, volumes and images), a log will be produced if and only if the verbose variable is enabled.

## Why is it important?
Support debugging certain failures while accessing the Docker engine

### DIFF
--- a/main.go
+++ b/main.go
@@ -323,11 +323,11 @@ func prune(cli *client.Client, deathNote *sync.Map) (deletedContainers int, dele
 		_ = try.Do(func(attempt int) (bool, error) {
 			argsClone := args.Clone()
 
-                  if verbose {
+			if verbose {
 				log.Printf("Deleting volumes with filter: %#v. (Attempt %d/%d)\n", argsClone, attempt, 10)
 			}
 
-      // API version >= v1.42 prunes only anonymous volumes: https://github.com/moby/moby/releases/tag/v23.0.0.
+			// API version >= v1.42 prunes only anonymous volumes: https://github.com/moby/moby/releases/tag/v23.0.0.
 			if serverVersion, err := cli.ServerVersion(context.Background()); err != nil && serverVersion.APIVersion >= "1.42" {
 				argsClone.Add("all", "true")
 			}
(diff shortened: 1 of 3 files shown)


HofmeisterAn (dismissed):
QQ: I don't remember anymore, did we decide to replace the CLI switches with environment variables or add the environment variables in addition to the switches?

mdelapenya:
We decided to use env vars, only

Explanation here: https://github.com/testcontainers/moby-ryuk/pull/49#pullrequestreview-1218678650

main.go
```diff
@@ -292,7 +323,11 @@ func prune(cli *client.Client, deathNote *sync.Map) (deletedContainers int, dele
 		_ = try.Do(func(attempt int) (bool, error) {
 			argsClone := args.Clone()
 
-			// API version >= v1.42 prunes only anonymous volumes: https://github.com/moby/moby/releases/tag/v23.0.0.
+      if verbose {
```
mdelapenya, line 326:
```suggestion
                  if verbose {
```

main.go
```diff
@@ -49,12 +52,14 @@ func newConfig(args []string) (*config, error) {
...
 		ConnectionTimeout:   60 * time.Second,
 		ReconnectionTimeout: 10 * time.Second,
+		Verbose:             false,
 	}
 
 	fs := flag.NewFlagSet("ryuk", flag.ExitOnError)
 	fs.SetOutput(os.Stdout)
 
 	fs.IntVar(&cfg.Port, "p", 8080, "Deprecated: please use the "+portEnv+" environment variable to set the port to bind at")
+	fs.BoolVar(&cfg.Verbose, "v", false, "If the Ryuk container should be started in verbose mode. If the "+verboseEnv+" environment variable is set, this flag will be ignored.")
```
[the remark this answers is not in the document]
mdelapenya: 🤔 IIRC I added for consistency with the other flags, but you are right, we should not add more. Will revert the flag here in this PR

main_test.go
```diff
@@ -361,6 +361,36 @@ func Test_newConfig(t *testing.T) {
...
+
+		config, err = newConfig([]string{})
+		require.Nil(t, err)
+		assert.False(t, config.Verbose)
+	})
+
+	t.Run("should set port from env with port flag and RYUK_VERBOSE environment variable", func(t *testing.T) {
+		t.Setenv(verboseEnv, "false")
+
+		config, err := newConfig([]string{"-v"})
```
HofmeisterAn, line 389:
```suggestion
		config, err := newConfig([]string{"-p", "3000"})
```

Should this be the port flag according to the test description:

> should set port from env with port flag and RYUK_VERBOSE environment variable

mdelapenya: I did not push a commit removing the test, as it's outdated 🤦 Pushed.

HofmeisterAn (approved):

PR #77, opened 2023-05-16
fix: Add filter all=true to volume prune

### LINKED ISSUE #75: Ryuk does not prune volumes (Moby >= v23.0.0, DD >= 4.19.0)
The latest release of Moby ([v23.0.0](https://github.com/moby/moby/releases/tag/v23.0.0)), which is included in Docker Desktop 4.19.0, introduces a breaking API change. It appears that the change is not "backwards compatible":

> API: Only anonymous volumes now pruned by default on API version >= v1.42. Pass the filter all=true to prune named volumes in addition to anonymous. https://github.com/moby/moby/pull/44259

To prune not just anonymous volumes, the filter `all=true` is required: [v1.42](https://docs.docker.com/engine/api/v1.42/#tag/Volume/operation/VolumePrune) vs. [v1.41](https://docs.docker.com/engine/api/v1.41/#tag/Volume/operation/VolumePrune). Maybe we can use the [NewVersionError](https://github.com/moby/moby/blob/3f7fde76c2e4e7c234e85c9a8634e2d7639f4f6d/client/errors.go#L61-L68) method to determine whether we need to include the additional filter or not. Following change fixes the issue for the latest version:

```diff
diff --git a/main.go b/main.go
index d59a3ee..fad8a74 100644
--- a/main.go
+++ b/main.go
@@ -290,6 +290,7 @@ func prune(cli *client.Client, deathNote *sync.Map) (deletedContainers int, dele
 		})
 
 		_ = try.Do(func(attempt int) (bool, error) {
+			args.Add("all", "true")
 			volumesPruneReport, err := cli.VolumesPrune(context.Background(), args)
 			for _, volumeName := range volumesPruneReport.VolumesDeleted {
 				deletedVolumesMap[volumeName] = true
```

We need to create a copy of `args` otherwise we include the filter in other prune operations too. Best we consider this for [images](https://github.com/testcontainers/moby-ryuk/blob/0f617fbc570c0bb64c27a8bd889d09df07f984f3/main.go#L306) too.

HofmeisterAn:
## What does this PR do?

The pull request checks Docker's API version. If the version is equal or greater than `1.42`, it adds `all=true` to the volume prune filter.

## Why is it important?

The latest Moby release ([v23.0.0](https://github.com/moby/moby/releases/tag/v23.0.0)), introduces a breaking API change. It is not backwards compatible. I checked an older version (Docker Desktop for Mac 4.14.1). The mentioned API version requires another filter (`all=true`) to prune all volumes. Without the additional filter, only anonymous volumes (those without names) are removed.

Do you have any better idea in mind to check the API version?

## Related issues

- Closes #75

### DIFF
--- a/main.go
+++ b/main.go
@@ -290,7 +290,14 @@ func prune(cli *client.Client, deathNote *sync.Map) (deletedContainers int, dele
 		})
 
 		_ = try.Do(func(attempt int) (bool, error) {
-			volumesPruneReport, err := cli.VolumesPrune(context.Background(), args)
+			argsClone := args.Clone()
+
+			// API version >= v1.42 prunes only anonymous volumes: https://github.com/moby/moby/releases/tag/v23.0.0.
+			if serverVersion, err := cli.ServerVersion(context.Background()); err != nil && serverVersion.APIVersion >= "1.42" {
+				argsClone.Add("all", "true")
+			}
+
+			volumesPruneReport, err := cli.VolumesPrune(context.Background(), argsClone)
 			for _, volumeName := range volumesPruneReport.VolumesDeleted {
 				deletedVolumesMap[volumeName] = true
 			}
@@ -303,8 +310,9 @@ func prune(cli *client.Client, deathNote *sync.Map) (deletedContainers int, dele
 		})
 
 		_ = try.Do(func(attempt int) (bool, error) {
-			args.Add("dangling", "false")
-			imagesPruneReport, err := cli.ImagesPrune(context.Background(), args)
+			argsClone := args.Clone()
+			argsClone.Add("dangling", "false")
+			imagesPruneReport, err := cli.ImagesPrune(context.Background(), argsClone)
 			for _, image := range imagesPruneReport.ImagesDeleted {
 				if image.Untagged != "" {
 					deletedImagesMap[image.Untagged] = true


main.go
```diff
@@ -303,8 +310,9 @@ func prune(cli *client.Client, deathNote *sync.Map) (deletedContainers int, dele
 		})
 
 		_ = try.Do(func(attempt int) (bool, error) {
-			args.Add("dangling", "false")
-			imagesPruneReport, err := cli.ImagesPrune(context.Background(), args)
+			argsClone := args.Clone()
```
mdelapenya, line 313: Good catch!

mdelapenya (approved):
LGTM!

main.go
```diff
@@ -290,7 +290,14 @@ func prune(cli *client.Client, deathNote *sync.Map) (deletedContainers int, dele
 		})
 
 		_ = try.Do(func(attempt int) (bool, error) {
-			volumesPruneReport, err := cli.VolumesPrune(context.Background(), args)
+			argsClone := args.Clone()
+
+			// API version >= v1.42 prunes only anonymous volumes: https://github.com/moby/moby/releases/tag/v23.0.0.
+			if serverVersion, err := cli.ServerVersion(context.Background()); err != nil && serverVersion.APIVersion >= "1.42" {
```
mdelapenya, line 296: I think it's fine at the moment there exists a proper comment with the changes

mdelapenya:
@HofmeisterAn do you think we should produce a `v0.4.1` patch release?

HofmeisterAn:
> @HofmeisterAn do you think we should produce a `v0.4.1` patch release?

Yes, then we can include it in the upcoming .NET release.

mdelapenya:
Sweet! Let's include also #71 in the new release

PR #78, opened 2023-05-17
fix: Invert the API version error check

HofmeisterAn:
## What does this PR do?

The PR inverts the API version error check. It was `err != nil` before.

## Why is it important?

Yesterday, while tidying up after my tests, I accidentally inverted the error check. The fix is not considered for the latest API.

## Related issues

- Relates #75


kiview (approved):

mdelapenya:
Oh I think that has been caused by #71, while resolving the conflicts with #77. My bad

PR #104, opened 2024-02-05
chore: include github actions in dependabot

mdelapenya:
## What does this PR do?
It adds GH actions to the dependabot manifest.

## Why is it important?
Keep dependencies updated.


HofmeisterAn (approved):
Ah, that is nice and convenient. I did not know about that. Thanks.

mdelapenya:
> Ah, that is nice and convenient. I did not know about that. Thanks.

Indeed, I noticed some errors regarding a deprecated nodejs runtime for the checkout action, and this will bring us the updates. In a follow-up PR I will pin the versions to a hash.

HofmeisterAn:
> > Ah, that is nice and convenient. I did not know about that. Thanks.
> 
> Indeed, I noticed some errors regarding a deprecated nodejs runtime for the checkout action, and this will bring us the updates. In a follow-up PR I will pin the versions to a hash.

Yes, I noticed them during the weekend in tc-dotnet too.

PR #110, opened 2024-02-23
Add Windows and more Linux platforms to the multi-arch image

gesellix:
Adds Windows and more Linux platforms to the multi-arch image.

This is a kind of minimal approach, optimizations and refactoring may be added later.

Follow-up of #40 
Clean version of https://github.com/gesellix/moby-ryuk/pull/2

Relates to https://github.com/testcontainers/moby-ryuk/issues/35
Relates to https://github.com/testcontainers/testcontainers-java/issues/5621

See https://github.com/gesellix/moby-ryuk/pkgs/container/moby-ryuk/182825956?tag=20240223.4 with a multi-arch image available for the proof of concept.
Docker image name: `ghcr.io/gesellix/moby-ryuk:20240223.4`

PR #120, opened 2024-03-01
Publish preview images, add checks for the multi-arch image variants

gesellix:
Follow up of #110


.github/workflows/build-docker-image.yml
```diff
@@ -74,26 +112,127 @@ jobs:
...
         with:
-          go-version-file: 'go.mod'
-      - run: go version
+          images: ${{ env.IMAGE_REPOSITORY }}
+          tags: |
+            type=sha,format=long
+          flavor: |
+            suffix=-windows.${{ matrix.os-version }}
+
+      - name: Login to Docker Hub
```
mdelapenya, line 126: We probably need to think about this feature twice: because it runs in a fork, the secrets are not available at the PR level.

Could you skip the docker login/push steps while in PRs?

gesellix: Yes, I'll try to add a check/flag during the next few days.

gesellix: I configured the jobs depending on the Docker Hub to only run on the default branch (main).

gesellix: ... oh wait. The 'build' job needs a special treatment.

gesellix: Would it make sense to add a "check" job to be required for pull requests, with the "check" depending on the "test-*" jobs? I'm going to change the "build" dependencies to be reduced to the "test-*" jobs, but I'm open to change the jobs and their structure like you prefer.

architectures.txt
```diff
@@ -0,0 +1,8 @@
+"386_linux_"
```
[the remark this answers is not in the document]
gesellix: Yes, that's a good idea. I could also rename the file to be more descriptive?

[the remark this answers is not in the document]
gesellix: "supported-architectures.txt"?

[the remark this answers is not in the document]
mdelapenya: Good and descriptive name 👍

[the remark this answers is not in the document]
gesellix: I've updated the filename and added minimal documentation.

gesellix:
As a thought: the workflows introduced here might be re-usable for other repositories in the "testcontainers" organisation. GitHub allows shared workflows. I would like to first _copy_ the workflows from here to one or two other repositories (see the pull request in the helloworld repo), so that we're able to decide whether an abstraction with shared workflows makes sense and how it should look like.

gesellix:
@mdelapenya did you have a chance to make a final review of the pull request? I would like to take the changes from here to the other pull request in the helloworld repository, so that we can continue working on WCOW support in the testcontainers-java repository.

HofmeisterAn:
> so that we can continue working on WCOW support in the testcontainers-java repository.

FYI @gesellix please be aware of the following limitations: https://github.com/testcontainers/testcontainers-dotnet/issues/1140#issuecomment-2006957120.

supported-architectures.txt
```diff
@@ -0,0 +1,8 @@
+"386_linux_"
+"amd64_linux_"
+"amd64_windows_10.0.17763.5458"
+"amd64_windows_10.0.20348.2322"
```
HofmeisterAn, line 4: Does it make sense to pin the actual Windows version here? This version will change; for example, `ltsc2022` is now `10.0.20348.2340` already.

gesellix: You're right, this won't be stable enough. I changed the check to only count the architectures and to not check the actual versions.

gesellix:
> > so that we can continue working on WCOW support in the testcontainers-java repository.
> 
> FYI @gesellix please be aware of the following limitations: [testcontainers/testcontainers-dotnet#1140 (comment)](https://github.com/testcontainers/testcontainers-dotnet/issues/1140#issuecomment-2006957120).

Thanks for the heads up. I think we can still continue working on WCOW support, because it's an improvement of the status quo without breaking existing use cases. Whether Testcontainers will declare the WCOW support as official or fully supported feature is probably a decision when all the building blocks are in place.

gesellix:
How and when are we going to continue here? I would like to improve the other tools, for example https://github.com/testcontainers/helloworld/pull/5, aiming at https://github.com/testcontainers/testcontainers-java/pull/8436 to become a step forward. Like already mentioned I would try to have some consistency across the different repositories (trying to minimize maintenance efforts), so this pull request could be a template for other repos. Consolidating them to a shared workflow would come after knowing all required variants.

If this is something you don't want to continue, feel free to leave a note. I guess WCOW support doesn't have a high priority. It would be great, though, if there would be some kind clarity in either working on WCOW support or explicitly not doing it.

mdelapenya (approved):
LGTM thanks!

mdelapenya:
@gesellix thanks for your work here, as always. It was delayed for no reason, so please forgive me if it took that long.

mdelapenya:
@gesellix the diff failed in the merge commit. Could you take a look please?

gesellix:
> @gesellix thanks for your work here, as always. It was delayed for no reason, so please forgive me if it took that long.

No reason to ask for forgiveness, my previous comment might look a bit harsh, but wasn't meant like that. I'm totally aware that maintaining many different projects is a lot of work and that sometimes less important aspects get out of sight :)

gesellix:
> @gesellix the diff failed in the merge commit. Could you take a look please?

Yes, I'll try to investigate asap (this week)!

gesellix:
@mdelapenya I found two issues, both have been fixed at #130 (see the details over there).

PR #123, opened 2024-03-14
chore: add caching to Linux docker run

stevenh:
Add cache mount directives to Linux docker container so that reruns are much quicker.


mdelapenya:
Hey @stevenh what are the motivation for this PR? Do you think it makes a real difference with the current state of the release process?

IIUC this will happen at build time of Ryuk, which is done just once per release. I see the benefits of building a Docker image faster, but want to check with you the implications/importance of this update.

Thanks!

stevenh:
For the release process no benefit. For the developer working on this it was much quicker iteration time, which was the motivation.

kiview (approved):
I think this is in general a good and modern `Dockerfile` / `docker build` approach and considered best practice.

mdelapenya (approved):
LGTM, thanks!

PR #130, opened 2024-05-23
Fix the manifest check

gesellix:
This is a follow-up to #120 

- The manifest now contains entries with "unknown" architecture. The entries aren't relevant to our check on existence of expected "linux" and "windows" entries, so we're simply ignoring "unknown" architectures.
- The diff has been fixed to use our input from stdin instead of trying to read any files.

### DIFF
--- a/.github/workflows/publish-docker-image.yml
+++ b/.github/workflows/publish-docker-image.yml
@@ -140,6 +140,8 @@ jobs:
 
       - name: Check multi-arch manifest
         run: |
+          cat architectures_actual.txt
           cat supported-architectures.txt | sort | uniq > architectures_expected.txt
           # we're counting the entries only, because windows os.version numbers change too frequently
-          diff $(cat architectures_expected.txt | wc -l) $(cat architectures_actual.txt | wc -l)
+          # exclude manifest entries with "unknown" architectures
+          diff <(echo "$(cat architectures_expected.txt | wc -l)") <(echo "$(cat architectures_actual.txt | grep -v unknown_unknown_ | wc -l)")


.github/workflows/publish-docker-image.yml
```diff
@@ -140,6 +140,8 @@ jobs:
 
       - name: Check multi-arch manifest
         run: |
+          cat architectures_actual.txt
           cat supported-architectures.txt | sort | uniq > architectures_expected.txt
           # we're counting the entries only, because windows os.version numbers change too frequently
-          diff $(cat architectures_expected.txt | wc -l) $(cat architectures_actual.txt | wc -l)
+          # exclude manifest entries with "unknown" architectures
+          diff <(echo "$(cat architectures_expected.txt | wc -l)") <(echo "$(cat architectures_actual.txt | grep -v unknown_unknown_ | wc -l)")
```
mdelapenya, line 147: So we want to just check the lines or the content? If the latter:
```suggestion
          diff <(echo "$(cat architectures_expected.txt | sort )") <(echo "$(cat architectures_actual.txt | grep -v unknown_unknown_ | sort)")
```

Thoughts?

gesellix: See the hint at the previous PR #120: for Windows, the `os.version` changes frequently, so comparing the content would make the comparison fragile. That's why checking only the count is used as indication whether all Linux and Windows architectures are included in the final image.

mdelapenya: Thanks for resolving my doubts 🙇

mdelapenya:
Mmmm failing again: https://github.com/testcontainers/moby-ryuk/actions/runs/9222172622/job/25373416643

@gesellix could you please take a look?

gesellix:
Yes, will do!

gesellix:
Oops - my bad: I only fixed the publish workflow, but not the build workflow. I'll create another pull request to fix the build workflow, too. Sorry.

gesellix:
I'll certainly extract a shared action so that the workflows become easier to maintain. This one and #131 are the best examples to have better abstractions and keeping code DRY 😅

gesellix:
That's great! Thanks! 🎉 
I"ll continue with the other repositories, stay tuned :)

PR #138, opened 2024-07-15
fix(gh-actions): depend on the right workflow

mdelapenya:
This PR is using the right job when releasing the Ryuk project.

It was possibly wrong after #120

### DIFF
--- a/.github/workflows/publish-docker-image.yml
+++ b/.github/workflows/publish-docker-image.yml
@@ -123,7 +123,7 @@ jobs:
   check-release:
     runs-on: ubuntu-latest
     needs:
-      - publish-multi-arch-image
+      - release
     steps:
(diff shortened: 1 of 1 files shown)


mdelapenya:
@gesellix I think we screwed it up in #120. Could you take a look 🙏 I'm planning a new release and it failed https://github.com/testcontainers/moby-ryuk/actions/runs/9942721870

Thank you in advance!

gesellix (approved):
Yes, I also think that's a leftover/copy&paste-error from the normal build workflow. The fix looks correct to me. 👍

gesellix:
@mdelapenya this one is also up-to-date now, including the fixed workflow reference: https://github.com/testcontainers/helloworld/pull/5

PR #141, opened 2024-07-27
feat: shutdown race resilience

stevenh:
A significant rewrite to ensure that we don't suffer from shutdown race conditions as the prune condition is met and additional resources are being created.

Previously this would remove resources that were still in use, now we retry if we detect new resources have been created within a window of the prune condition triggering.

This supports two new environment configuration settings:
- RYUK_REMOVE_RETRIES - The number of times to retry removing a resource.
- RYUK_REQUEST_TIMEOUT - The timeout for any Docker requests.
- RYUK_RETRY_OFFSET - The offset added to the start time of the prune pass that is used as the minimum resource creation time


mdelapenya:
For some reason, I need to set up the loadConfig tests with `RYUK_VERBOSE=false` in order to make the tests pass locally:

```diff
tests := map[string]struct {
		setEnv   func(*testing.T)
		expected config
	}{
		"defaults": {
+			setEnv: func(t *testing.T) {
+				t.Helper()
+				t.Setenv("RYUK_VERBOSE", "false")
+			},
			expected: config{
				ConnectionTimeout:   time.Minute,
				Port:                8080,
				ReconnectionTimeout: time.Second * 10,
				RemoveRetries:       10,
				RequestTimeout:      time.Second * 10,
				RetryOffset:         -time.Second,
				ShutdownTimeout:     time.Minute * 10,
			},
		},
		"custom": {
			setEnv: func(t *testing.T) {
				t.Helper()
				t.Setenv("RYUK_PORT", "1234")
				t.Setenv("RYUK_CONNECTION_TIMEOUT", "2s")
				t.Setenv("RYUK_RECONNECTION_TIMEOUT", "3s")
				t.Setenv("RYUK_REQUEST_TIMEOUT", "4s")
				t.Setenv("RYUK_REMOVE_RETRIES", "5")
				t.Setenv("RYUK_RETRY_OFFSET", "-6s")
				t.Setenv("RYUK_SHUTDOWN_TIMEOUT", "7s")
+				t.Setenv("RYUK_VERBOSE", "false")
			},
			expected: config{
				Port:                1234,
				ConnectionTimeout:   time.Second * 2,
				ReconnectionTimeout: time.Second * 3,
				RequestTimeout:      time.Second * 4,
				RemoveRetries:       5,
				RetryOffset:         -time.Second * 6,
				ShutdownTimeout:     time.Second * 7,
			},
		},
	}
```

A bug in the env library for default booleans?

mdelapenya (commented):
Thanks for opening this @stevenh. I see a lot of changes here, which I'd suggest splitting in different PRs:

- config refactor
- run function

Ryuk is a critical piece in the Testcontainers ecosystem, so updating it should be done very carefully to not affect all the related projects.

Would you mind separating this big PR into multiple ones? I can help you out with it if needed.

Thanks!

stevenh:
> For some reason, I need to set up the loadConfig tests with `RYUK_VERBOSE=false` in order to make the tests pass locally:
> 
> A bug in the env library for default booleans?

Interesting what's the error you get?

stevenh:
> For some reason, I need to set up the loadConfig tests with `RYUK_VERBOSE=false` in order to make the tests pass locally:
> 
> snip...
> 
> A bug in the env library for default booleans?

If you had environment vars set then the test would have failed, I've updated to clean the current environment when it starts which could fix the issue you saw.

mdelapenya:
Hi @stevenh, thanks for this PR, I've started the review and I see that there are some changes that would deserve its own PR:
- the addition of a library to read the env vars. Is there any other library you considered for that? I like this one is simple and with zero deps.
- the bumps for the different Go versions and tools (golangci-lint)
- the real refactor

Could you separate them into three? I can help out with the two first so you can focus on the real meat here

stevenh:
Ok so little more involved due to go bump needed new linter so we need to merge in the following order:
1. https://github.com/testcontainers/moby-ryuk/pull/160
2. https://github.com/testcontainers/moby-ryuk/pull/159 -- Also includes version bump so it complies
3. https://github.com/testcontainers/moby-ryuk/pull/141 -- Will need a rebase after the config merge

mdelapenya:
Given #159 has been merged, could you please rebase this one? 🙏

main_test.go
```diff
@@ -1,314 +0,0 @@
-package main
```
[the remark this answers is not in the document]
stevenh: I think each implementation of TC should test its own reaper behaviour, which I suspect they already do, `testcontainers-go` does.

I do think we should prove docker  works how the mock expects, that's a miss and we should add that back in before merging to make sure there's nothing stilly wrong.

I'd be tempted to do that without the testcontainers-go dependency though, thoughts?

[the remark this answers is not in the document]
mdelapenya: > I do think we should prove docker works how the mock expects, that's a miss and we should add that back in before merging to make sure there's nothing stilly wrong. I'd be tempted to do that without the testcontainers-go dependency though, thoughts?

At the moment we are testing against real containers, I'm fine. I chose tc-go as the means to create those containers in order to be a consumer of the library. But we could just simply use `docker run` plus labels if we want to.

README.md
```diff
@@ -1,43 +1,71 @@
...
 
-- `RYUK_CONNECTION_TIMEOUT` - Environment variable that defines the timeout for Ryuk to receive the first connection (default: 60s). Value layout is described in [time.ParseDuration](https://golang.org/pkg/time/#ParseDuration) documentation.
-- `RYUK_PORT` - Environment variable that defines the port where Ryuk will be bound to (default: 8080).
-- `RYUK_RECONNECTION_TIMEOUT` - Environment variable that defines the timeout for Ryuk to reconnect to Docker (default: 10s). Value layout is described in [time.ParseDuration](https://golang.org/pkg/time/#ParseDuration) documentation.
-- `RYUK_VERBOSE` - Environment variable that defines if Ryuk should print debug logs (default: false).
+The following environment variables can be configured to change the behaviour:
+
+| Environment Variable        | Default | Format  | Description  |
+| --------------------------- | ------- | ------- | ------------ |
+| `RYUK_CONNECTION_TIMEOUT`   | `60s`   | [Duration](https://golang.org/pkg/time/#ParseDuration) | The duration without receiving any connections which will trigger a shutdown |
```
mdelapenya, line 64: This table is neat!

mdelapenya (dismissed):
LGTM, thanks!

I like the changes about the vocabulary: I'll miss the ryuk names, although I admit that using the names and words in this PR is now more inclusive and more maintainable for the long term, so thanks for this.

Once merged, as we are building the images for the commit, we can immediately test this image in testcontainers-go, using the commit-sha as docker tag (see https://hub.docker.com/r/testcontainers/ryuk/tags)

stevenh:
@mdelapenya added some docker cli based tests for container, network, image and volume reaping so should be good to go.

reaper_test.go
```diff
@@ -0,0 +1,726 @@
...
+
+	require.Contains(t, log.String(), "shutdown timeout")
+}
+
+func TestReapContainer(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
+	t.Cleanup(cancel)
+
+	// Run a test container.
+	args := []string{"container", "run", "--detach", "--quiet"}
```
mdelapenya, line 554: What do you prefer: 1) shelling out to the binary, 2) using the docker client from the docker/docker dependency (we already have it), or 3) keeping testcontainers-go as dogfooding, showcasing the benefits of using it?

mdelapenya (commented):
One comment regarding shelling out to the docker binary.

Other than that, will take the final view this weekend for final approval 🙏  I'm doing conferences next week, so I'll try to do before that.

reaper_test.go
stevenh: I wanted to avoid using testcontainers-go as that creates a circular repository dependency, which has bitten me in the past.

Good call on client though, simpler to use except for the image, which isn't too bad.

stevenh: He says and then ContainerCreate doesn't pull missing images :(

mdelapenya (approved):
LGTM, great PR, thanks for adding clarity in the reap process.

Cheers!

PR #159, opened 2024-09-07
chore: simply config

stevenh:
Simplify the configuration so its easier to add new settings, clearer what the default values are and which environment variables impact the running of container.

This includes compatibility with slog for future use.


main.go
```diff
@@ -25,89 +22,18 @@ import (
...
 )
 
-type config struct {
-	Port                int
-	ConnectionTimeout   time.Duration
-	ReconnectionTimeout time.Duration
-	Verbose             bool
-}
-
-// newConfig parses command line flags and returns a parsed config. config.timeout
```
mdelapenya, line 49: We are losing the command flags support in this PR, right? Is this intended?

mdelapenya: Have you considered using https://github.com/ardanlabs/conf/blob/master/doc.go?

stevenh: I couldn't see how that could have ever been used so didn't see any point in keeping. Correct me if I'm wrong?

mdelapenya: Yeah, this is a valid concern. @eddumelendez @kiview what do you think? @stevenh raises a valid concern here about removing the flags support: are the Ryuk flags used anywhere else? If used in any other project, external to testcontainers, are we fine removing that support, keeping the env vars only?

From my point of view, the simplest, the best, so removing that support would reduce the codebase.

eddumelendez: I don't think any issue removing the flags

config.go
```diff
@@ -0,0 +1,52 @@
...
+
+	// Port is the port to listen on for connections.
+	Port uint16 `env:"RYUK_PORT" envDefault:"8080"`
+
+	// Verbose is whether to enable verbose aka debug logging.
+	Verbose bool `env:"RYUK_VERBOSE" envDefault:"false"`
+}
+
+// LogAttrs returns the configuration as a slice of attributes.
+func (c config) LogAttrs() []slog.Attr {
```
mdelapenya, line 33: We are not using this method. Do we want to do it on startup if verbose mode is enabled?

mdelapenya (approved):
LGTM, although I'd like to double check with you the need of logging the config right after it's loaded (if verbose)

main.go
kiview: As long as we don't break integration with other Testcontainers implementations, it is fine 👍

config.go
stevenh: Follow up will use it, didn't want to also change the logger in this PR.

main.go
stevenh: If they are using the container, all is good.

mdelapenya (approved):
LGTM, thanks!

PR #160, opened 2024-09-07
chore: bump versions

stevenh:
Bump go, modules and golangci-lint versions so they are compatible with incoming changes.


.github/workflows/golangci-lint.yml
```diff
@@ -17,10 +17,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
-        with:
-          go-version-file: 'go.mod'
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@3cfe3a4abbb849e10058ce4af15d205b6da42804 # v4
+        uses: golangci/golangci-lint-action@v6
```
mdelapenya, line 21: Can we use the pinned commit for v6? 🙏 It's recommended for not finding surprises in undesired bumps

go.mod
```diff
@@ -1,21 +1,21 @@
 module github.com/testcontainers/moby-ryuk
 
-go 1.21
+go 1.23
 
 require (
-	github.com/docker/docker v27.1.1+incompatible
+	github.com/docker/docker v27.2.0+incompatible
```
mdelapenya, line 6: I'm fine with this bump, although there are some conversations with other maintainers regarding the default docker version, as it could limit the features Ryuk could provide regarding the docker versions in the workers in which the different tc-libs run.

Not a blocker though. 

Cc/ @kiview

mdelapenya (dismissed):
LGTM just a few informative comments. 

Thanks!

.github/workflows/golangci-lint.yml
stevenh: I can but I tend to take opposite view, trust developers to keep compatibility in major version allowing you to get backwards compatible fixes and enhancements as they are released, reducing the maintenance overhead and increasing readability. 

This is also inline with the other use directives above.

Thoughts?

go.mod
stevenh: Also as a general rule I find updating regularly causes less pain and keeps on top of any security issues.

I believe the main docker release is now following semantic versioning and moving towards module support (can't come soon enough), but I'm sure you know more than I do on that 😄

.github/workflows/golangci-lint.yml
mdelapenya: It's a practice I try to follow regarding security: https://www.stepsecurity.io/blog/pinning-github-actions-for-enhanced-security-a-complete-guide

stevenh: > It's a practice I try to follow regarding security: https://www.stepsecurity.io/blog/pinning-github-actions-for-enhanced-security-a-complete-guide

I suspected that might be the case, updated.

mdelapenya (approved):
LGTM, thanks!

We can inform consumers about the bump of the docker version, and let them decide when to bump to a new Ryuk version.

PR #163, opened 2024-09-30
feat: run tests on windows

mdelapenya:
It enables the tests on Windows workers


mdelapenya:
Converting to draft: the GH workers run in non-WSL windows, so we need to run them in the Docker infra, with a self-hosted Windows box. I will update this PR once we have access.

PR #167, opened 2024-10-09
fix: use Started instead of listening

mdelapenya:
## What does this PR do?
It uses `Started` instead of `listening` in the log to define that Ryuk properly initialised.

## Why is this important?
To avoid breaking existing clients of Ryuk, which could be updating the Ryuk version in the properties file

## Related issues
- Relates https://github.com/testcontainers/testcontainers-java/issues/9369

## Follow-up questions
I wonder what are the different use cases to have a different ryuk version at the properties level and not use the one provided by the library.

I acknowledge that we introduced a breaking change when we updated the log format to a more structured one, which we are fixing in this PR, but if the user puts a wrong version in the properties, like a very outdated ryuk version, what can we do in that case?

### DIFF
--- a/reaper.go
+++ b/reaper.go
@@ -140,7 +140,9 @@ func newReaper(ctx context.Context, options ...reaperOption) (*reaper, error) {
 		return nil, fmt.Errorf("listen: %w", err)
 	}
 
-	r.logger.Info("Started", fieldAddress, r.listener.Addr().String())
+	// This log message, in uppercase, is in use in different Testcontainers libraries,
+	// so it is important to keep it as is to not break the current behavior of the libraries.
+	r.logger.Info("Started client processing")
 
 	return r, nil
 }
--- a/reaper.go
+++ b/reaper.go
@@ -142,7 +142,7 @@ func newReaper(ctx context.Context, options ...reaperOption) (*reaper, error) {
 
 	// This log message, in uppercase, is in use in different Testcontainers libraries,
 	// so it is important to keep it as is to not break the current behavior of the libraries.
-	r.logger.Info("Started client processing")
+	r.logger.Info("Started", fieldAddress, r.listener.Addr().String())
 
 	return r, nil
 }
--- a/reaper.go
+++ b/reaper.go
@@ -140,7 +140,7 @@ func newReaper(ctx context.Context, options ...reaperOption) (*reaper, error) {
 		return nil, fmt.Errorf("listen: %w", err)
 	}
 
-	r.logger.Info("listening", fieldAddress, r.listener.Addr().String())
+	r.logger.Info("Started", fieldAddress, r.listener.Addr().String())
 
 	return r, nil
 }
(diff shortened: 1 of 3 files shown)


eddumelendez (dismissed):

reaper.go
```diff
@@ -140,7 +140,7 @@ func newReaper(ctx context.Context, options ...reaperOption) (*reaper, error) {
 		return nil, fmt.Errorf("listen: %w", err)
 	}
 
-	r.logger.Info("listening", fieldAddress, r.listener.Addr().String())
+	r.logger.Info("Started", fieldAddress, r.listener.Addr().String())
```
stevenh, line 143: suggestion: While it's listening here its not processing yet, I would update the following instead and also add a comment as to why the upper case so someone doesn't change at a later date because its inconsistent with everything else which is lower case.:
```go
	r.logger.Info("client processing started")
```

stevenh (changes requested):

reaper.go
mdelapenya: But the regular expression that tc-java uses is not camelcase, it expects `.*Started.*`. That's why I added the simple `Started` word, to avoid weird combinations including lowercase and upper case words, like this one:

```go
	r.logger.Info("client processing Started")
```

I'm adding the following comment:
```go
	// This log message, in uppercase, is in use in different Testcontainers libraries,
	// so it is important to keep it as is to not break the current behavior of the libraries.
	r.logger.Info("Started", fieldAddress, r.listener.Addr().String())
```

stevenh: You could swap it round and say:
```go
r.logger.Info("Started client processing")
```

The main point was to have it just before it would actually be able to process requests but it in practice it should be fine either way.

reaper.go
```diff
@@ -140,7 +140,9 @@ func newReaper(ctx context.Context, options ...reaperOption) (*reaper, error) {
 		return nil, fmt.Errorf("listen: %w", err)
 	}
 
-	r.logger.Info("listening", fieldAddress, r.listener.Addr().String())
+	// This log message, in uppercase, is in use in different Testcontainers libraries,
+	// so it is important to keep it as is to not break the current behavior of the libraries.
+	r.logger.Info("Started client processing")
```
stevenh, line 145: Sorry @mdelapenya I think you might not have understood what I meant in my previous suggestion.

I was suggesting we change the [log entry in processClients](https://github.com/testcontainers/moby-ryuk/blob/main/reaper.go#L187) instead of this one, as that is just before the service is ready to actually accept clients instead of this one when only the listener is enabled.

I wasn't suggesting changing the content of message here.

So to be clear two options:
1. Change the message here to what you had e.g. `r.logger.Info("Started", fieldAddress, r.listener.Addr().String())`
2. Change the message in `processClients` r.logger.Info("Started client processing")

In either case should keep the output of the address as thats the only way the user can see the address that the service is listening on if a ephemeral port is requested.

stevenh (changes requested):

reaper.go
mdelapenya: I'm really sorry @stevenh. Did read it very vertically in the middle of the conference, my fault.

mdelapenya: @stevenh I think `1.` is better, thanks for pointing out

stevenh (approved):

PR #169, opened 2024-10-11
fix: slow shutdown after signal

stevenh:
Fix shutdown after signal to allow the last client disconnect to trigger a prune check. Previous this would always wait for the shutdown timeout, which by default is 10 minutes, resulting in an unnecessarily slow shutdown.


mdelapenya (approved):
LGTM, thanks!

PR #170, opened 2024-10-11
fix: alpine image version

stevenh:
Fix the alpine image version using both 3.19 and 3.20 which could lead to compatibility issues.


mdelapenya (approved):
LGTM

PR #173, opened 2024-10-22
fix: multiple filters

stevenh:
Fix multiple filters which was only applying the removal of the last iterated set.

Add tests to validate that:
* Multiple filters are correctly processed
* Duplicate filters are correctly ignored
* Immediate shutdown occurs if no clients
    
This increasing code coverage to 89.4%.

Use best effort in removal if resource changes are detected after a shutdown has been signalled to allow as much to be cleaned up as possible.

This includes a new setting `ChangesRetryInterval` exposed by the env variable `RYUK_CHANGES_RETRY_INTERVAL` and defaults to 1 second which is used to control the interval between retries if resource changes are detected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ printf "label=something_else" | nc -N localhost 8080
 In the ryuk window you'll see containers/networks/volumes deleted after 10s
 
 ```log
-time=2024-09-30T19:42:30.000+01:00 level=INFO msg=starting connection_timeout=1m0s reconnection_timeout=10s request_timeout=10s shutdown_timeout=10m0s remove_retries=10 retry_offset=-1s port=8080 verbose=false
+time=2024-09-30T19:42:30.000+01:00 level=INFO msg=starting connection_timeout=1m0s reconnection_timeout=10s request_timeout=10s shutdown_timeout=10m0s remove_retries=10 retry_offset=-1s changes_retry_interval=1s port=8080 verbose=false
 time=2024-09-30T19:42:30.001+01:00 level=INFO msg="Started"
 time=2024-09-30T19:42:30.001+01:00 level=INFO msg="client processing started"
 time=2024-09-30T19:42:38.002+01:00 level=INFO msg="client connected" address=127.0.0.1:56432 clients=1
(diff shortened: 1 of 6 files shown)


reaper.go
```diff
@@ -416,17 +430,19 @@ func (r *reaper) affectedContainers(since time.Time, args filters.Args) ([]strin
...
-			return nil, fmt.Errorf("container %s: %w", container.ID, errChangesDetected)
+			errChanges = append(errChanges, fmt.Errorf("container %s: %w", container.ID, errChangesDetected))
+			continue
 		}
 
 		containerIDs = append(containerIDs, container.ID)
 	}
 
-	return containerIDs, nil
+	return containerIDs, errors.Join(errChanges...)
```
mdelapenya, line 440: IIUC, we are returning the error chain here, which means that all resources created after the prune init time will contribute an error here. And that error chain will contribute to the one in the `resources` method, but because they will be `errChangesDetected`, the pruneWait function will continue processing them as a log.Warn (see reaper.go:L329.

If this is assumption correct, then it's fine not returning a nil error but the error chain. Can you confirm that?

mdelapenya (dismissed):
LGTM, I added one comment just to check I'm understanding the error chain well.

reaper.go
stevenh: There's a few impacts from this:
1. The returned won't be logged by the resources method
2. The error will contribute to the error chain and passed up
3. `puneWait` will retry if `errChangesDetected` is detected and shutdown timeout hasn't been reached

stevenh: Off back off this question I added code to ensure that shutdown timeout is enforced, triggering best effort clean up.

Let me know this makes sense?

config.go
```diff
@@ -28,6 +28,9 @@ type config struct {
 	// calculated time will trigger a retry to ensure in use resources are not removed.
 	RetryOffset time.Duration `env:"RYUK_RETRY_OFFSET" envDefault:"-1s"`
 
+	// ChangesRetryInterval is the interval between retries if resource changes are detected.
+	ChangesRetryInterval time.Duration `env:"RYUK_CHANGES_RETRY_INTERVAL" envDefault:"1s"`
```
mdelapenya, line 32: We need to add this new setting to the readme. We must produce a minor for the new release then

mdelapenya (dismissed):

config.go
mdelapenya: Question: do we want to be configurable, or just opinionated and enforce a fixed value?

stevenh: I considered that, I went with just having everything in the config for consistency, happy to revisit in a future PR.

README.md
```diff
@@ -68,13 +68,14 @@ time=2024-09-30T19:42:52.216+01:00 level=INFO msg=done
...
+| `RYUK_PORT`                   | `8080`  | `uint16` | The port to listen on for connections |
+| `RYUK_RECONNECTION_TIMEOUT`   | `10s`   | [Duration](https://golang.org/pkg/time/#ParseDuration) | The duration after the last connection closes which will trigger resource clean up and shutdown |
+| `RYUK_REQUEST_TIMEOUT`        | `10s`   | [Duration](https://golang.org/pkg/time/#ParseDuration) | The timeout for any Docker requests |
+| `RYUK_REMOVE_RETRIES`         | `10`    | `int` | The number of times to retry removing a resource |
+| `RYUK_RETRY_OFFSET`           | `-1s`   | [Duration](https://golang.org/pkg/time/#ParseDuration) | The offset added to the start time of the prune pass that is used as the minimum resource creation time. Any resource created after this calculated time will trigger a retry to ensure in use resources are not removed |
+| `RYUK_CHANGES_RETRY_INTERVAL` | `1s`    | [Duration](https://golang.org/pkg/time/#ParseDuration) | The the interval between retries if resource changes are detected |
```
mdelapenya, line 79: Typo:
```suggestion
| `RYUK_CHANGES_RETRY_INTERVAL` | `1s`    | [Duration](https://golang.org/pkg/time/#ParseDuration) | The interval between retries if resource changes are detected |
```

mdelapenya: Can we also explain it with more details:

> The internal between retries if resource changes (containers, networks, images, and volumes) are detected while pruning resources

mdelapenya (approved):
LGTM; just a suggestion in the docs

Thanks!

mdelapenya (approved):
LGTM, thanks!

PR #176, opened 2024-11-20
docs: document how to integrate with Ryuk

mdelapenya:
## What does this PR do?
It describes the integration points to be implemented by Testcontainers libraries in order to communicate with Ryuk.

## Why is it important?
Better documentation is always good

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,3 +79,31 @@ The following environment variables can be configured to change the behaviour:
 | `RYUK_CHANGES_RETRY_INTERVAL` | `1s`    | [Duration](https://golang.org/pkg/time/#ParseDuration) | The internal between retries if resource changes (containers, networks, images, and volumes) are detected while pruning |
 | `RYUK_VERBOSE`                | `false` | `bool` | Whether to enable verbose aka debug logging |
 | `RYUK_SHUTDOWN_TIMEOUT`       | `10m`   | [Duration](https://golang.org/pkg/time/#ParseDuration) | The duration after shutdown has been requested when the remaining connections are ignored and prune checks start |
+
+## Integrating Ryuk in the Testcontainers Libraries
+
+The Testcontainers libraries can be configured to use Ryuk to remove resources after a test session has completed.
+
+- Identify test session semantics for the Testcontainers library. For example, a test session could be a single test method, a test class, or a test suite. As reference, please consider taking a look at Go's implementation [here](https://golang.testcontainers.org/features/test_session_semantics/). This unique identifier for the test session semantic, is referenced as `SESSION_ID` from now on.
+- Use the above configuration to start Ryuk as a special container within the library. For that, read the above environment variables and/or from the Testcontainers properties file, which is located in the home directory of the user. Regarding precedence, the environment variables must have higher precedence than the properties file.
+    - Define Ryuk as a container with privileged access.
+    - Define a wait strategy for the listening port, defined by the `RYUK_PORT` environment variable. This is necessary to ensure that Ryuk is ready to receive messages from the Testcontainers library.
+    - Bind the Docker socket to the Ryuk container, so that it can communicate with the Docker daemon. This is necessary to be able to create and remove resources.
+        - E.g. `${RESOLVED_DOCKER_SOCKET}:/var/run/docker.sock`, where `${RESOLVED_DOCKER_SOCKET}` is the path to the Docker socket discovered by the Testcontainers library.
+    - Optionally, name the container using the `SESSION_ID` to make it easier to identify the container in the Docker daemon.
+    - Expose the port defined by the `RYUK_PORT` environment variable, so that the Testcontainers library can send messages to Ryuk.
+    - Add a special label to the Ryuk container in order to avoid removing it by mistake.
+        - E.g. `org.testcontainers.reaper=true`, `org.testcontainers.ryuk=true`, etc.
+        - If you already use a specific label for reaping resources, please remember to remove it from the Ryuk container for the same reason.
+    - Ryuk should run in the default bridge network of the Docker runtime.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -79,3 +79,31 @@ The following environment variables can be configured to change the behaviour:
...
+    - Optionally, name the container using the `SESSION_ID` to make it easier to identify the container in the Docker daemon.
+    - Expose the port defined by the `RYUK_PORT` environment variable, so that the Testcontainers library can send messages to Ryuk.
+    - Add a special label to the Ryuk container in order to avoid removing it by mistake.
+        - E.g. `org.testcontainers.reaper=true`, `org.testcontainers.ryuk=true`, etc.
+        - If you already use a specific label for reaping resources, please remember to remove it from the Ryuk container for the same reason.
+    - Ryuk should run in the default bridge network of the Docker runtime.
+- Every time a Docker resource is created in the Testcontainers library, Ryuk must be informed about it. This can be done by sending a message to Ryuk with the Docker labels of the resource, as a set of key-value pairs. In general, it's a good practice to always send the same set of labels for all the resources, including the above `SESSION_ID`, so that Ryuk can consistenly identify and remove the created resources after the test session has completed.
```
HofmeisterAn, line 99: **question**: Are you really doing this for every resource? .NET sends it only once at the start and then assigns the same value and label the resources.

HofmeisterAn: In .NET, we follow these steps:

1. Create and start Ryuk.
2. Connect to Ryuk.
3. Send the filter.
4. Maintain the connection.

Then we use the ID from the filter and label every resource we create with it.

HofmeisterAn: Sorry, I do not fully understand. .NET establishes only one connection to Ryuk. Each test process has its own instance. Each test process creates Ryuk, connects, sends the filter, and maintains the connection only once for all running tests (within the process).

stevenh: To clarify @HofmeisterAn do you have an explicit setup step you use for tests then in .NET which handles the connection setup an tear down, instead of a container or network creation automatically handling that?

HofmeisterAn: Ah, when I mentioned .NET, I meant Testcontainers for .NET. Sorry for the confusion. I was referring to

> Every time a Docker resource is created in the Testcontainers library, Ryuk must be informed about it.

and just wanted to point out that the .NET implementation doesn’t send the filter for every resource.

stevenh: Yep I understood you meant Testcontainers .NET. What I'm trying to understand is what in testcontainers .NET is responsible for setting up ryuk and connecting too it. You seemed to infer it was something outside of just creating a container with testcontainers .NET, is that the case?

How testcontainers-go works is every resource created checks to ensure ryuk is running, if not it creates it and in either way connects to it, so ryuk know there is an additional dependency.

ryuk monitors these connections and when the last one disconnects, it runs the clean up. This means that the clean up should be quick as its a triggered event but also it means that if there is an issue along the way and something triggers an unexpected shutdown the failing connections would still trigger a clean up.

For the go implementation this is import as the test infrastructure has a global timeout and if that triggers it doesn't gracefully clean up test resources.

In the wider use, a test container could be used for a single test in a suite, so having each container do this validation helps to ensure ryuk and and orphaned resources only run while needed.

HofmeisterAn: > You seemed to infer it was something outside of just creating a container with testcontainers .NET, is that the case?

No, we start Ryuk with the first container resource.

> How testcontainers-go works is every resource created checks to ensure ryuk is running, if not it creates it and in either way connects to it, so ryuk knows there is an additional dependency.

This part is slightly different in .NET. Every resource we create checks if Ryuk is running. If it is not, we create it; if it is, we skip the creation. We do not establish another connection to Ryuk. The connection is created with the first resource that creates Ryuk (the resource does not hold the connection; it is held by the test process).

stevenh: So to confirm, if there was a test that ran for a period of time after all docker resources were unneeded, these would be keep available as its not until the process exits that the connection is removed?

Example in sudo code
```
func testWithContainer() {
    container = tc.Run("myimage"....)
    // Do things with container....
}

func main() {
      testWithoutContainer()
      testWithContainer()
      // Resources not cleaned up yet...
      moreTestsWithoutContainer() 
      // Connection to ryuk still open, container not cleaned up until main exits?
}
```

README.md
```diff
@@ -79,3 +79,31 @@ The following environment variables can be configured to change the behaviour:
...
+
+## Integrating Ryuk in the Testcontainers Libraries
+
+The Testcontainers libraries can be configured to use Ryuk to remove resources after a test session has completed.
+
+- Identify test session semantics for the Testcontainers library. For example, a test session could be a single test method, a test class, or a test suite. As reference, please consider taking a look at Go's implementation [here](https://golang.testcontainers.org/features/test_session_semantics/). This unique identifier for the test session semantic, is referenced as `SESSION_ID` from now on.
+- Use the above configuration to start Ryuk as a special container within the library. For that, read the above environment variables and/or from the Testcontainers properties file, which is located in the home directory of the user. Regarding precedence, the environment variables must have higher precedence than the properties file.
+    - Define Ryuk as a container with privileged access.
+    - Define a wait strategy for the listening port, defined by the `RYUK_PORT` environment variable. This is necessary to ensure that Ryuk is ready to receive messages from the Testcontainers library.
```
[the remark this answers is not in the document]
stevenh: Yes the configuring any `RYUK_` prefixed environment variables is optional, including `RYUK_PORT`, if however the user specifies it, it should be honoured, so we ensure connections are made to that port.

README.md
HofmeisterAn: > Connection to ryuk still open, container not cleaned up until main exits?

In theory, that is true, but .NET provides a concept for releasing unmanaged resources: [Dispose](https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose). Typically, when a test completes, the `Dispose` method cleans up the container if it is no longer needed. Ryuk serves as a fallback, guaranteeing cleanup in cases when the test process crashes.

stevenh: Yep that's the same with testcontainers-go, its a fall back for tests written correctly, so having an extra delay is not a big deal.

However documenting that you can use a connection per resource to ensure timely clean up is still of benefit IMO, thoughts?

PR #211, opened 2025-09-01
feat: use scratch as base image

mdelapenya:
## What does this 
- **deps: bump Alpine build image to 3.22**
- **feat: use scratch as base image**

## Why is it important?
Reduce the attach surface, running Ryuk in an image that contains nothing.

### DIFF
--- a/linux/Dockerfile
+++ b/linux/Dockerfile
@@ -21,12 +21,18 @@ RUN --mount=type=cache,target=/go/pkg/mod \
     go build -ldflags '-s' -o /bin/ryuk
 
 # -----------------
-# Distributed Image
+# Certificates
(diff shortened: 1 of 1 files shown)


kiview (approved):
Makes we wonder why we were running with an Alpine based one for the so long 🙂

mdelapenya:
@kiview I found this: https://github.com/testcontainers/helloworld/pull/2, which comes from https://github.com/testcontainers/testcontainers-java/issues/3317

It makes me wonder if we should roll this back 🤔 In tc-go, the `WaitForPort` strategy allows skipping internal checks, so users can avoid the error seen in those issues.

@eddumelendez thoughts?

PR #212, opened 2025-09-01
chore: reduce binary size with upx

mdelapenya:
## What does this PR do?
It uses [upx](https://upx.github.io/) to optimise the final Go binary.

I explored it on https://github.com/testcontainers/helloworld, and applied all the knowledge from that here.

## Why is it important?
Instead of producing a Docker image with 10MB, we are producing one with ~5MB

### DIFF
--- a/linux/Dockerfile
+++ b/linux/Dockerfile
@@ -32,7 +32,10 @@ RUN --mount=type=cache,target=/go/pkg/mod \
     -o /bin/ryuk .
 
 # Compress with UPX (trade-off: smaller size vs startup time)
-RUN apk add --no-cache upx && upx --best --lzma /bin/ryuk
+# Note: UPX is not available for s390x architecture
+RUN if [ "$(uname -m)" != "s390x" ]; then \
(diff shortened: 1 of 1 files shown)


kiview (commented):
Do you have numbers on startup impact?

Context, I believe for normal users (i.e. not CI), smaller images are less important than fast startup times. Depends on the accumulated values, but my hunch is, we don't want to make this trade.

kiview (approved):
I've let Copilot crunch the numbers on this in https://github.com/testcontainers/moby-ryuk/pull/213.

The TLDR is (which I admittedly might trust to blindly, but isn't that the name of the game nowadays? 😅):
> UPX provides net benefits when: `(Pull Time Savings × Pulls) > (Startup Overhead × Starts)`
> 
> With 3.9 second pull savings vs 0.1ms startup overhead (100-iteration average), UPX is beneficial in virtually all real-world scenarios where containers are pulled more than once per 39,000 starts.

Given that we see Ryuk being pulled up to 19M times per month, we end up with real world savings of **95 TB per month**!

